PR #38, opened 2026-03-17
fix(receiver/windowseventtrace): Always shut down session

mrsillydog:
### Proposed Change

Currently in the windowseventtrace receiver, ownership of the session is ambiguous - the `logsReceiver` struct starts the session in logsReceiver.Start(), but after the session is passed down to the Consumer, responsibility falls on the Consumer to shut down the session in Consumer.Stop().

However, if any errors occur after session initialization, but before Consumer initialization, then Consumer.Stop never gets passed to the stop queue and the session is therefore never stopped. Example errors that could cause this case are below:

```golang
// session is running
	failedProviders := totalProviders - successfulProviders
	if lr.cfg.RequireAllProviders && failedProviders > 0 {
		lr.logger.Error("Failed to enable all providers",
			zap.String("session", lr.cfg.SessionName),
			zap.Int("successfulProviders", successfulProviders),
			zap.Int("failedProviders", failedProviders))
// HERE -->
		return
	}

	eventConsumer := etw.NewRealTimeConsumer(ctx, lr.logger, lr.session, lr.cfg.Raw)
	err := eventConsumer.Start(ctx)
	if err != nil {
		lr.logger.Error("Failed to start ETW consumer", zap.Error(err))
// HERE -->
		return
	}
// this needs to be called in order for the session to be shut down and isn't if we return early
	lr.stopQueue = append(lr.stopQueue, func(ctx context.Context) error { return eventConsumer.Stop(ctx) })
```

To fix this, this PR adjusts the session ownership to be fully within the purview of the logsReceiver. Instead of calling session.controller.Stop via the proxy of Consumer.Stop, it is instead called directly from the logsReceiver.Shutdown method.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


mrsillydog:
Moved from https://github.com/observIQ/bindplane-otel-collector/pull/3273

mrsillydog:
This PR has been successfully tested. How I did so:

Config:
```yaml
windowseventtrace/s-01KKF14X5A1107FDPDA9JRXNRS__logs:
    providers:
        - level: informational
          name: Microsoft-Windows-Kernel-File
        - level: informational
          name: Fake-Prov-Name // this provider does not exist
    raw: false
    require_all_providers: false // this is the key flag
    session_buffer_size: 64
    session_name: Bindplane-ETW-Session-Leak
```

Reproduction:
- Run a collector based on main with the above config.
- Verify that a session is running: `logman query "Bindplane-ETW-Session-Leak" -ets`
- Shut down the collector: `Stop-Service -Name observiq-otel-collector`
- Check to ensure that the session has also been shut down: `logman query "Bindplane-ETW-Session-Leak" -ets` ✅ 
- Reconfigure the collector to set `require_all_providers` to `true`
- Start the collector
- Verify a session is running
- Stop the collector
- Verify the session is still running ❌ 

Fix:
- Run a collector based on this branch with the above config.
- Verify that a session is running: `logman query "Bindplane-ETW-Session-Leak" -ets`
- Shut down the collector: `Stop-Service -Name observiq-otel-collector`
- Check to ensure that the session has also been shut down: `logman query "Bindplane-ETW-Session-Leak" -ets` ✅ 
- Reconfigure the collector to set `require_all_providers` to `true`
- Start the collector
- Verify a session is running
- Stop the collector
- Verify the session has been shut down ✅

schmikei (approved):
Seems to work pretty consistently :)

PR #52, opened 2026-03-18
chore: Add make targets for more OS/ARCH build combos

mrsillydog:
### Proposed Change

- Create a build target for each supported BDOT OS/ARCH combo
- Ensure each target can build successfully in CI

This also allows for simple building of a new binary based on local changes for an arbitrary distro.

##### Checklist
- [X] Changes are tested
- [X] CI has passed


jsirianni:
Does this need to run in CI? We already run tests for all three platforms; this should prove that the code can compile.

I do like have the ability to target a platform for local builds.

Dylan-M (approved):

mrsillydog:
> Does this need to run in CI? We already run tests for all three platforms; this should prove that the code can compile.
> 
> I do like have the ability to target a platform for local builds.

@jsirianni The original reasoning behind adding the collector build in CI was to ensure that the V1 collector would still compile with the PR's version of the components. My thought was that if we're validating this, we shouldn't just be validating it for one specific platform. Our CI in BDOT would re-run the build for each platform upon component changes, so I think it makes sense to do the same here. opentelemetry-collector-contrib also [builds for each platform](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f0bac736c6598e6dce6160c55ccba989ab396a26/.github/workflows/build-and-test-windows.yml#L41) when running CI.

jsirianni (approved):

PR #56, opened 2026-03-18
feat(receiver/windowseventtrace): Set log.record.original on attributes

mrsillydog:
### Proposed Change

This PR allows users to configure their Windows Event Trace receiver to include log.record.original as an attribute. This is functionally a way to include both raw and parsed data in one log. This data is not included for raw logs, as it's simply a duplicate.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


mrsillydog:
Tested:
include_log_body_original = true, raw = false places the contents of Raw into a log.record.original attribute
include_log_body_original = false, raw = false has no attributes
include_log_body_original = true, raw = true has no attributes

schmikei (approved):
Love matching with the https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/windowseventlogreceiver/README.md thanks for doing this :)

PR #58, opened 2026-03-18
fix(chronicleexporter): correctly report logs dropped metric

colelaven:
### Proposed Change
Currently the Refused Spans metric has always been reported as 0 when reporting metrics to the Chronicle APIs.

This PR:
- Accurately tracks and reports the Logs Dropped (Refused Spans) metric in the chronicle exporter when a permanent error occurs.

Note - This metric will only be imported into GCP when there is a permanent error hitting the chronicle import logs endpoint, but the exporter can still successfully hit the import metrics endpoint. So a completely invalid / incorrect project number, customerID, credentials won't result in this metric getting imported.

The easiest way I found to replicate this is to use credentials that have permission to importStatsEvent but not to importLogs. I have a service account `cole-dropped-logs-testing` configured like this that can be used for testing.

image

### DIFF
--- a/exporter/chronicleexporter/grpc_exporter.go
+++ b/exporter/chronicleexporter/grpc_exporter.go
@@ -216,6 +216,11 @@ func (exp *grpcExporter) uploadToChronicle(ctx context.Context, request *api.Bat
 			exp.telemetry.ExporterRequestCount.Add(ctx, 1,
 				metric.WithAttributeSet(attribute.NewSet(attrErrorUnknown)))
 
+			if exp.metrics != nil {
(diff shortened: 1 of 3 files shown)


shazlehu (approved):
Graph looks good [here](https://console.cloud.google.com/monitoring/dashboards/builder/56450109-144e-426b-8cf6-4a1614a61674;duration=PT30M?project=telemetry-sandbox-340915) using a telemetry generator running at 1/s:

Image

Important consideration since this a `delta` metric is that you need to set the interval correctly (in this case 5m) to get the graph to appear correctly. At the default value (1m), the rate is 5/s.

Image

Image

PR #101, opened 2026-03-24
feat: REST API Header parameter

Caleb-Hurshman:
### Proposed Change

Add 2 new parameters to the REST API receiver to accept a map of additional headers, one for secure headers that will be masked, and one for any other headers that don't need to be secured.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/restapireceiver/config.go
```diff
@@ -341,6 +353,33 @@ func (c *Config) Validate() error {
...
+		if err := validateHeaderValue(string(value)); err != nil {
+			return fmt.Errorf("invalid sensitive header value for %q: %w", name, err)
+		}
+	}
+
+	// Check for duplicate header names across headers and sensitive_headers
+	for name := range c.SensitiveHeaders {
+		if _, exists := c.Headers[name]; exists {
+			return fmt.Errorf("header %q is defined in both headers and sensitive_headers; use one or the other", name)
+		}
```
Dylan-M, line 401: The overlap check here is case-sensitive, but HTTP headers are case-insensitive — `req.Header.Set` canonicalizes keys under the hood. So something like `headers: {"x-auth": "foo"}` + `sensitive_headers: {"X-Auth": "bar"}` would slip through validation but collide at request time.

Same thing applies within a single map too — `{"X-Foo": "a", "x-foo": "b"}` is effectively a duplicate that won't get caught.

Could use `http.CanonicalHeaderKey` to normalize before comparing, both across maps and within each one. Would be good to add a test for the case-insensitive collision as well.

receiver/restapireceiver/config.go
```diff
@@ -341,6 +353,33 @@ func (c *Config) Validate() error {
...
+		if err := validateHeaderValue(string(value)); err != nil {
+			return fmt.Errorf("invalid sensitive header value for %q: %w", name, err)
+		}
+	}
+
+	// Check for duplicate header names across headers and sensitive_headers
+	for name := range c.SensitiveHeaders {
+		if _, exists := c.Headers[name]; exists {
+			return fmt.Errorf("header %q is defined in both headers and sensitive_headers; use one or the other", name)
```
Dylan-M: Minor nit: the README says "Must not overlap with `headers`" but this error says "use one or the other". The error message reads better — maybe update the README to match?

receiver/restapireceiver/config.go
```diff
@@ -341,6 +353,33 @@ func (c *Config) Validate() error {
...
+		}
+	}
+
+	// Validate sensitive headers
+	for name, value := range c.SensitiveHeaders {
+		if err := validateHeaderName(name); err != nil {
+			return fmt.Errorf("invalid sensitive header name %q: %w", name, err)
+		}
+		if err := validateHeaderValue(string(value)); err != nil {
+			return fmt.Errorf("invalid sensitive header value for %q: %w", name, err)
```
Dylan-M, line 373: Minor nit: the header name shows up in validation errors for `sensitive_headers`. The value is excluded which is good, but is the name itself ever considered sensitive? Something like `X-Secret-Internal-Service-Token` could reveal infrastructure details. Probably fine, just wanted to flag it.

Dylan-M (changes requested):

receiver/restapireceiver/client.go
```diff
@@ -125,10 +125,13 @@ func (c *defaultRESTAPIClient) GetJSON(ctx context.Context, requestURL string, p
 		return nil, fmt.Errorf("failed to apply authentication: %w", err)
 	}
 
-	// Set headers
+	// Set default headers
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("Content-Type", "application/json")
```
Dylan-M: Not from this PR, but setting `Content-Type` on a GET with no body is technically incorrect per HTTP semantics. Now that custom headers make this more visible, might be worth a follow-up to drop it or only set it when there's a body.

receiver/restapireceiver/config.go
Dylan-M: I would suggest getting Tony's thoughts, but the poor guy is so overloaded right now :(

receiver/restapireceiver/config.go
Caleb-Hurshman: Reworded the README to be more explicit

receiver/restapireceiver/config.go
Caleb-Hurshman: My instinct says no, but open to the idea that there are edge cases. We don't treat the field names for the other auth modes as sensitive (i.e., `header_name` for API Key auth).

Dylan-M: Perfect, just something clade flagged to me and made me think about it.

receiver/restapireceiver/config.go
```diff
@@ -341,6 +354,35 @@ func (c *Config) Validate() error {
...
+		}
+	}
+
+	// Check for duplicate header names across headers and sensitive_headers (case-insensitive)
+	for name := range c.SensitiveHeaders {
+		for hName := range c.Headers {
+			if http.CanonicalHeaderKey(name) == http.CanonicalHeaderKey(hName) {
+				return fmt.Errorf("header %q is defined in both headers and sensitive_headers; use one or the other", name)
+			}
```
Dylan-M: The cross-map check looks good now, but I think we still need the same treatment within each map individually. Something like `headers: {"X-Foo": "a", "x-foo": "b"}` is effectively a duplicate that would slip through since Go map keys are case-sensitive but HTTP headers aren't. Could we please add the same `http.CanonicalHeaderKey` comparison within each map as well?

Dylan-M (changes requested):

Dylan-M (approved):
Looks good — all feedback addressed. Nice work on the case-insensitive handling.

PR #105, opened 2026-03-25
fix(`restapireceiver`): Improve response timestamp handling

Caleb-Hurshman:
### Proposed Change

Improves how the REST API receiver handles the timestamps from the response returned by the API. 
- If a user has a timestamp format configured, try to parse that format first
- If there's no configured format, or the configured format fails to parse (API returns a different format than it expects), fallback to a regex for common ISO8601 and RFC 3339 formats

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


schmikei (approved):
Nice work! Tested around with a variety of formats and appreciate the resiliency with the regex.

PR #123, opened 2026-03-26
chore: Harden Get() resiliency in Pebble Extension

schmikei:
<!-- ## Important (read before submitting)
In order for changes to be captured in changelog correctly please add one of the following prefixes to the title. **Note** the parenthesis are optional and so is any text in them.
- `feat(OPTIONAL):` = New features
- `fix(OPTIONAL):` = Bug fixes
- `deps(OPTIONAL):` = Dependency updates, primarily dependabot
-->

### Proposed Change

We have not seen this in the field but theoretically the API by pebble just returns sequential memory that pebble manages which means another program can modify its contents. 

Because of this we should perform a copy like how bbolt and badger do:


https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1a616d98873ca26622ef6d4698c2e7d1ddd7a25f/extension/storage/filestorage/client.go#L116-L121

```go
value := bucket.Get([]byte(op.Key))                                                                                                                                                                                                      
  if value != nil {                                                                                                                                                                                                                        
      op.Value = make([]byte, len(value))                                                                                                                                                                                                  
      copy(op.Value, value)                                                                                                                                                                                                                
  }
```

We do this for badger already:

https://github.com/observIQ/bindplane-otel-contrib/blob/c3593fba6f74f37364d600d742ba5adaa0238345/extension/badgerextension/internal/client/client.go#L143-L151

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Dylan-M (approved):
Clean and correct. Nice test too.

mrsillydog (approved):

PR #135, opened 2026-03-31
feat: send structured JSON for map-body logs in azureloganalyticsexporter

eKuG:
### Proposed Change

When raw_log_field is not set, logs with a map body now have each key emitted as a top-level JSON field (mapping directly to Sentinel table columns). Logs with a string/empty body are wrapped in {"RawData": ...} so ingestion still works for unstructured data. Metadata fields (TimeGenerated, SeverityText, SeverityNumber, TraceId, SpanId) are included in both paths.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -84,15 +84,59 @@ func (m *azureLogAnalyticsMarshaler) transformLogsToSentinelFormat(ctx context.C
...
+				if ts.IsZero() {
+					ts = logRecord.ObservedTimestamp().AsTime()
+				}
+				if ts.IsZero() {
+					ts = time.Now()
+				}
+				entry["TimeGenerated"] = ts.Format(time.RFC3339)
+
+				// Severity
+				entry["SeverityText"] = logRecord.SeverityText()
```
Caleb-Hurshman, line 131: Is it possible that the metadata fields we set here (severity text, traceid, spanid, etc.) could collide with fields on the map body? If so, those fields will get overwritten and lost.
Either way, this would be a good test to add, and good to document this behavior.

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -84,15 +84,59 @@ func (m *azureLogAnalyticsMarshaler) transformLogsToSentinelFormat(ctx context.C
 	if m.cfg.RawLogField != "" {
 		return m.transformRawLogsToAzureLogAnalyticsFormat(ctx, ld)
 	}
-	td := plogotlp.NewExportRequestFromLogs(ld)
+	var outputSlice []map[string]interface{}
```
Caleb-Hurshman: This initializes to nil, so if there are no log records, json.Marshal(outputSlice) will return null instead of [].
It looks like we aren't checking for null anywhere up the chain, so we'll attempt to send null to Azure.
```suggestion
	outputSlice := make([]map[string]interface{}, 0)
```
It depends on what Azure expects, but this could save us some error responses if we ever encounter an empty `ld`

Caleb-Hurshman (commented):
These changes mostly look good, I just have a couple of questions.

Caleb-Hurshman (approved):

PR #136, opened 2026-03-31
fix: classify HTTP errors for retry in azureloganalyticsexporter

eKuG:
### Proposed Change

Add error classification so exporterhelper correctly handles permanent vs transient errors. Permanent errors (400, 401, 403) are now marked with consumererror.NewPermanent to stop futile retries. Transient errors (429, 502, 503, 504) remain retryable, with Retry-After header support via NewThrottleRetry. Marshal errors are also marked permanent. Azure SDK built-in retry is preserved for first-tier fast retries.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Caleb-Hurshman (approved):
Looks good ✅

PR #138, opened 2026-03-31
feat(restapireceiver): Add end time parameters

Caleb-Hurshman:
### Proposed Change

The REST API receiver had no way to put an upper bound on the time range that it queried. This PR adds 2 fields to the timestamp configuration:
- `end_timestamp_param_name`
- `end_timestamp_value`

Edit: this PR now also updates the receiver to included a time bound for any request, not just those using timestamp pagination. The top level start time parameter takes the place of the timestamp pagination parameter, using that value as an initial timestamp when timestamp pagination is configured.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


receiver/restapireceiver/README.md
```diff
@@ -123,14 +135,13 @@ Use `auth_mode: none` for public APIs that don't require authentication. No addi
 
 #### Timestamp-Based Pagination
 
-| Field                                       | Type   | Default | Required | Description                                                                                                                                                                                                                          |
-| ------------------------------------------- | ------ | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `pagination.timestamp.param_name`           | string |         | `true`   | Query parameter name for timestamp (e.g., "t0", "since", "after", "start_time")                                                                                                                                                      |
```
eKuG, line 130: `pagination.timestamp.param_name`, `pagination.timestamp.timestamp_format`, and `pagination.timestamp.initial_timestamp` are removed and replaced with field? Any existing user configs using the old field names will silently stop working (the old fields are simply ignored via mapstructure). Is this intentional?

receiver/restapireceiver/pagination.go
```diff
@@ -71,24 +71,24 @@ func newPaginationState(cfg *Config) *paginationState {
 		}
 
 	case paginationModeTimestamp:
-		// Set initial timestamp if provided, otherwise start from zero time.
+		// Set initial timestamp from the top-level start_time_value if provided.
 		// Config validation ensures the timestamp is parseable.
-		if cfg.Pagination.Timestamp.InitialTimestamp != "" {
-			if isEpochFormat(cfg.Pagination.Timestamp.TimestampFormat) {
+		if cfg.StartTimeValue != "" && cfg.StartTimeValue != "now" {
```
eKuG: Does this skip the "now" case entirely, leaving CurrentTimestamp as zero time? For timestamp pagination, a zero timestamp means "start from the beginning."

receiver/restapireceiver/pagination.go
```diff
@@ -497,6 +514,24 @@ func parseTimestampString(s string) (time.Time, bool) {
 	return time.Time{}, false
 }
 
+// formatTimeBoundValue formats a time-bound value string for use as a query parameter.
+// If value is "now", returns the current time formatted per the given format.
+// Otherwise, returns the value as-is (it was already validated by Config.Validate()).
+func formatTimeBoundValue(value, format string) string {
```
eKuG, line 549: I guess this handles `now`, but timestamp pagination path doesn't, it relies on `CurrentTimestamp` which was never set. Should we set `CurrentTimestamp = time.Now().UTC()` when the value is "now"?

eKuG:
Shoulw we add a case to check end_time > start_time always?

eKuG (changes requested):

receiver/restapireceiver/README.md
Caleb-Hurshman: Hmm, it's possible to make a custom unmarshalling for the old fields, I'll see what I can do

receiver/restapireceiver/pagination.go
```diff
@@ -158,53 +160,70 @@ func buildPaginationParams(cfg *Config, state *paginationState) url.Values {
...
 		// No pagination parameters
 	}
 
+	// Add time-bound parameters for non-timestamp pagination modes.
+	// For timestamp pagination, start time is handled above (it advances through data).
+	if cfg.Pagination.Mode != paginationModeTimestamp {
+		if cfg.StartTimeParamName != "" && cfg.StartTimeValue != "" {
+			params.Set(cfg.StartTimeParamName, formatTimeBoundValue(cfg.StartTimeValue, cfg.TimestampFormat))
+		}
+	}
```
eKuG, line 233: For offset/limit or page_size pagination, buildPaginationParams is called per-page. If `start_time_value` is "now", each page gets a different start time, which shifts the query window mid-pagination.

Recommendation : Resolve "now" once per poll cycle

Caleb-Hurshman: Good catch, just pushed a change

eKuG (approved):
LGTM

PR #141, opened 2026-04-02
feat(restapireceiver): Support NDJSON return format

Caleb-Hurshman:
### Proposed Change

Not every API returns a nice JSON object, some do newline delimited JSON. In those cases, pagination metadata will often be returned in a completely separate response, or in the response header.

This PR adds 2 new configuration options:
- specify the returned data format (json or ndjson)
- specify where to extract pagination metadata (body or header) 

Tested these changes against a mock api modeled after the Akamai SIEM API that returns ndjson, using token pagination, with the pagination metadata in a separate response (mock api source code and receiver config available upon request).

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "mcp__claude_ai_Linear__get_issue",
-      "mcp__claude_ai_Linear__list_comments"
-    ]
-  }
-}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "mcp__claude_ai_Linear__get_issue",
+      "mcp__claude_ai_Linear__list_comments"
+    ]
+  }
+}
--- a/receiver/restapireceiver/client.go
+++ b/receiver/restapireceiver/client.go
@@ -43,12 +43,12 @@ type restAPIClient interface {
 	// Returns an array of map[string]any representing the JSON objects.
 	GetJSON(ctx context.Context, requestURL string, params url.Values) ([]map[string]any, error)
 	// GetFullResponse fetches the full JSON response from the specified URL.
-	// Returns the full response as map[string]any for pagination parsing.
-	GetFullResponse(ctx context.Context, requestURL string, params url.Values) (map[string]any, error)
+	// Returns the full response as map[string]any for pagination parsing, plus response headers.
+	GetFullResponse(ctx context.Context, requestURL string, params url.Values) (map[string]any, http.Header, error)
 	// GetNDJSON fetches an NDJSON response from the specified URL.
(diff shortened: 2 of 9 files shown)


.claude/settings.local.json
eKuG, line 1: Does this file need to be committed?

Caleb-Hurshman: Nope, just another thing that's not gitignored in this repo yet.

receiver/restapireceiver/client.go
```diff
@@ -234,39 +238,149 @@ func (c *defaultRESTAPIClient) GetFullResponse(ctx context.Context, requestURL s
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return data, metadata, resp.Header, nil
+}
+
+// parseNDJSON parses an NDJSON response body into data objects and a metadata object.
+// The last non-empty line is treated as metadata; all preceding lines are data objects.
+// Empty lines are skipped.
+func parseNDJSON(body []byte, logger *zap.Logger) ([]map[string]any, map[string]any, error) {
```
eKuG: - If an API returns NDJSON with no metadata line (just data), the last data record gets silently eaten, is it?
- Is this possible?

Caleb-Hurshman: That comment is a little outdated. That is the case when the source field is set as 'body'. If that's not what the API does, then the receiver is misconfigured. Usually any API that does cursor/token pagination should return pagination metadata no matter what.

receiver/restapireceiver/receiver.go
```diff
@@ -189,15 +190,70 @@ func (b *baseReceiver) saveCheckpoint(ctx context.Context) error {
...
+	// response headers and inject them into the metadata map. This lets the
+	// pagination logic find them through the same field-name lookup it uses for body
+	// attributes — no changes needed downstream.
+	if b.cfg.Pagination.ResponseSource == responseSourceHeader {
+		if metadata == nil {
+			metadata = make(map[string]any)
+		}
+		for _, fieldName := range b.paginationResponseFields() {
+			if headerVal := respHeaders.Get(fieldName); headerVal != "" {
+				metadata[fieldName] = headerVal
```
eKuG, line 227: does parsePaginationResponse do a type assertion like val.(float64)

Caleb-Hurshman: Yes, see pagination.go:263 for an example

eKuG: When response_source: header, header values are injected as strings. So metadata["total_record_count"] would be "4" (a string), not float64(4) or int(4). Neither type assertion matches a string. Would this be the case at any point in time?

Caleb-Hurshman: Ahh I see what you're getting at. Yes, I think that would fail silently. I'll update the behavior 👍

eKuG (approved):

PR #171, opened 2026-04-03
feat: add a byte batcher processor

schmikei:
### Proposed Change

This PR introduces a new bytebatcher processor that batches telemetry before the next component in the pipeline, similar in role to the core [batch processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor), but with a different batching signal.

The batch processor primarily bounds batches by item count and timeout. bytebatcher bounds batches by an estimated serialized size of each payload (using pdata’s protobuf marshaling sizer as a proxy for “size on the wire”) and by a configurable flush interval. Whichever threshold is hit first triggers a flush.

Some notes before full enablement and I'm still testing:

If the downstream consumer returns an error on flush, the batch is not retried (to avoid duplicate export); the failure is currently logged and the batch is dropped. I haven't seen this on my initial testing but this component needs more backpressure testing

Defaults:

- `bytes`: 1MiB
- `flush_interval`: 1s

Validated in local/testing environments; additional load and backpressure testing are planned

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


dyl10s:
I spent some time yesterday looking into this. From my initial understanding, this is using the same measuring strategy as the `exporterhelper`. I will put some time on your calendar when you get back from vacation so you can help me understand the goals of this processor.

schmikei:
Closing as unplanned for now

PR #173, opened 2026-04-06
feat(azureblobpollingreceiver): add glob support and blob format config

DanQUlXOTE:
### Proposed Change
Adds two enhancements to the Azure Blob Polling Receiver:

**Glob support for `root_folder`**
- `root_folder` now accepts glob patterns (e.g., `linux/*`, `logs-*/app`) to match
multiple virtual directory prefixes
- Uses Azure's hierarchy listing API to resolve glob patterns into concrete prefixes at
poll time
- Non-glob values continue to work as before with no behavior change

**`blob_format` config for NDJSON and raw text parsing**
- New `blob_format` config option with three values: `otlp` (default), `json` (NDJSON),
and `text` (raw text)
- `json` format parses newline-delimited JSON, creating one log record per line
- `text` format creates one log record per blob with the raw content as the body
- Adds `NDJSONLogsConsumer` and `TextLogsConsumer` to the `blobconsume` package
- Fixes a bug where failed `FromRaw` calls in the NDJSON consumer could leave orphaned
empty records in the output

##### Checklist
- [x] Changes are tested
- [x] CI has passed


internal/azureblob/blob_client.go
```diff
@@ -99,6 +122,14 @@ func (a *AzureClient) StreamBlobs(ctx context.Context, container string, prefix
 			return
 		}
 
+		pageNumber++
+		blobsInPage := len(resp.Segment.BlobItems)
+		totalStreamed += blobsInPage
+		a.logger.Info("Azure API page received",
```
Caleb-Hurshman: Between this and all the other logging added in this PR, these receivers will become a lot noisier. This could/should be a debug level log.

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -308,47 +332,119 @@ func (r *pollingReceiver) pullBlobs(ctx context.Context, startingTime, endingTim
...
+
+	if len(matched) == 0 {
+		r.logger.Warn("Glob pattern matched zero directories, no blobs will be scanned",
+			zap.String("root_folder", r.cfg.RootFolder),
+			zap.String("static_prefix", staticPrefix),
+			zap.Int("candidates", len(prefixes)))
+		return []string{}
+	}
+
+	r.logger.Info("Glob expanded root_folder",
```
Caleb-Hurshman: Another log message that could be debug level.

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -83,13 +87,30 @@ func newLogsReceiver(id component.ID, logger *zap.Logger, cfg *Config, nextConsu
 	}
 
 	r.supportedTelemetry = pipeline.SignalLogs
-	r.consumer = blobconsume.NewLogsConsumer(nextConsumer)
+
+	switch cfg.BlobFormat {
+	case BlobFormatJSON:
+		logger.Info("Using NDJSON blob format consumer")
```
Caleb-Hurshman: All 3 of these can be debug level

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -416,16 +517,21 @@ blobLoop:
...
 		}
 	}
 
 	r.wg.Wait()
 
-	if err := r.makeCheckpoint(ctx); err != nil {
-		r.logger.Error("Error while saving checkpoint", zap.Error(err))
-	}
+	processed := int(processedBlobCount.Load())
+	r.logger.Info("Batch processing summary",
```
Caleb-Hurshman: This was a debug log before, we should keep it that way.

internal/blobconsume/text_consumer_test.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func Test_rawTextLogsConsumer_EmptyContent(t *testing.T) {
+	sink := &consumertest.LogsSink{}
+	con := NewRawTextLogsConsumer(sink)
+
+	err := con.Consume(context.Background(), []byte(""))
+	require.NoError(t, err)
+
+	require.Equal(t, 1, sink.LogRecordCount())
```
Caleb-Hurshman: The NDJSON test returns 0 for the test where `Consume(context.Background(), []byte(""))` is called. This should probably match that pattern so we don't send empty logs.

internal/blobconsume/ndjson_consumer.go
```diff
@@ -0,0 +1,97 @@
...
+	scopeLogs := resourceLogs.ScopeLogs().AppendEmpty()
+	logRecords := scopeLogs.LogRecords()
+
+	now := pcommon.NewTimestampFromTime(time.Now())
+	lines := bytes.Split(entityContent, []byte("\n"))
+	skipped := 0
+
+	for i, line := range lines {
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 {
```
Caleb-Hurshman, line 58: This is what I mentioned in text_consumer_test.go, the NDJSON consumer properly skips empty lines.

receiver/azureblobpollingreceiver/README.md
```diff
@@ -1,3 +1,7 @@
+> [!WARNING]
+> **This component has been migrated to [bindplane-otel-contrib](https://github.com/observiq/bindplane-otel-contrib/tree/main/receiver/azureblobpollingreceiver).**
+> This module is retained for reference and will be removed after September 2026.
```
Caleb-Hurshman: This warning is no longer needed in this repo.

receiver/azureblobpollingreceiver/README.md
```diff
@@ -69,6 +73,62 @@ This prevents duplicate processing of blobs and ensures data continuity across c
...
+
+| Format | Description                                                                                                                                                                          |
+| ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `otlp` | (Default) OTLP JSON format. Blobs are unmarshaled using the standard OpenTelemetry `plog.JSONUnmarshaler`. Use this when blobs were written by the Azure Blob Exporter.             |
+| `json` | Newline-delimited JSON (NDJSON). Each line is parsed as a JSON object and becomes a separate log record with the parsed object as the body. Malformed lines are skipped with a warning. |
+| `text` | Raw text. The entire blob content is set as the body of a single log record.                                                                                                        |
+
+`json` and `text` are only supported on **logs** pipelines. Metrics and traces pipelines only support `otlp`.
```
Caleb-Hurshman: It would be nice to check this value against `telemetry_type` in the Validate() method to catch this earlier.

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -308,47 +332,119 @@ func (r *pollingReceiver) pullBlobs(ctx context.Context, startingTime, endingTim
...
+	}
+
+	staticPrefix, globPattern := splitGlobPrefix(r.cfg.RootFolder)
+
+	prefixes, err := r.azureClient.ListPrefixes(ctx, r.cfg.Container, staticPrefix)
+	if err != nil {
+		r.logger.Warn("Failed to list prefixes for glob expansion, falling back to literal root_folder",
+			zap.String("root_folder", r.cfg.RootFolder),
+			zap.Error(err))
+		return []string{r.cfg.RootFolder}
```
Caleb-Hurshman: To clarify, this will just return the literal string representation of the glob pattern? This seems like it will not be useful for matching any real glob, unless it's common to have a glob named something like `"linux/*"`.

Caleb-Hurshman: We should either return an error here, skip the poll, or fall back to static_prefix instead.

Caleb-Hurshman (changes requested):

internal/azureblob/blob_client.go
DanQUlXOTE: Changed to Debug

receiver/azureblobpollingreceiver/receiver.go
DanQUlXOTE: Changed to Debug

receiver/azureblobpollingreceiver/receiver.go
DanQUlXOTE: Changed to Debug

receiver/azureblobpollingreceiver/receiver.go
DanQUlXOTE: Changed to Debug

internal/blobconsume/text_consumer_test.go
DanQUlXOTE: updated

receiver/azureblobpollingreceiver/receiver.go
DanQUlXOTE: Changed to Static_prefix fallback

receiver/azureblobpollingreceiver/README.md
DanQUlXOTE: Checking in Validate() now

Caleb-Hurshman (approved):
LGTM ✅

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -308,47 +344,121 @@ func (r *pollingReceiver) pullBlobs(ctx context.Context, startingTime, endingTim
...
+
+	if !isGlobPattern(r.cfg.RootFolder) {
+		return []string{r.cfg.RootFolder}
+	}
+
+	staticPrefix, globPattern := splitGlobPrefix(r.cfg.RootFolder)
+
+	prefixes, err := r.azureClient.ListPrefixes(ctx, r.cfg.Container, staticPrefix)
+	if err != nil {
+		r.logger.Warn("Failed to list prefixes for glob expansion, falling back to static prefix",
```
Caleb-Hurshman: Let's bump this up to an error log to make it more visible. With NSG containers, this fallback could get expensive. Consider making this fallback opt-in to avoid users getting caught off guard by this behavior.

receiver/azureblobpollingreceiver/path_parser.go
```diff
@@ -72,8 +72,11 @@ func parseWithPlaceholders(blobPath, pattern string) (*time.Time, error) {
 		i += end + 1
 	}
 
-	// Compile and match the regex
-	re, err := regexp.Compile("^" + regexPattern)
+	// Compile and match the regex. The pattern is intentionally not anchored
```
Caleb-Hurshman: This is good for NSG logs, but we should protect against this when the user doesn't have a root_folder.
If the root_folder trimming matches something, we can relax the parsing.
If it didn't, add the anchor back.

Caleb-Hurshman (commented):
A couple of smaller comments

receiver/azureblobpollingreceiver/receiver.go
DanQUlXOTE: Good call - both points addressed in 517d016b.

- Log level bumped from `Warn` to `Error` on the ListPrefixes failure path.
- Fallback is now opt-in via a new `fallback_on_glob_failure` config (default `false`). When left at the default, a ListPrefixes failure logs the error and the poll lists nothing for that cycle, so we don't silently fall back to scanning under a broad static prefix. Set to `true` to keep the previous fallback behavior.

README updated to document the new field and the rationale.

receiver/azureblobpollingreceiver/path_parser.go
DanQUlXOTE: Done in 517d016b.

`parseTimeFromPattern` now takes an `anchored bool`. `trimMatchedRoot` returns `(string, bool)` indicating whether a root prefix was actually stripped. In `shouldProcessBlob`, we pass `anchored = !rootTrimmed`:

- When a `root_folder` (or glob expansion of it) matches and is trimmed, the pattern is matched unanchored against the remainder — this is the NSG case.
- When nothing was trimmed (no `root_folder` configured, or no match), the pattern keeps the original `^` anchor, so we don't accidentally match the time segment in arbitrary positions in the path.

New test covers both: NSG-style path with trimmed root parses successfully, and the same shape with no root configured rejects with `does not match pattern`.

Caleb-Hurshman (approved):
Awesome ✅ 
Thanks for iterating

PR #174, opened 2026-04-06
feat(`restapireceiver`): Add config fingerprint to checkpoint

Caleb-Hurshman:
### Proposed Change

Update the REST API receiver's checkpointing logic to use a config fingerprint to track changes to the configuration to determine whether to use the checkpoint values or not.
The driver of this was a user running the receiver with storage enabled, then configuring a new `initial_timestamp` to limit the data, but not having that `initial_timestamp` be picked up because of the existing checkpoint.

Only certain fields are used to compute the checkpoint fingerprint:
- url
- initial_timestamp
- pagination_mode

This PR will need an update once #138 is merged (Edit: Updated)

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/restapireceiver/receiver.go
```diff
@@ -147,7 +191,33 @@ func (b *baseReceiver) adjustPollInterval(result pollResult) {
...
+// receiver fetches. When any of these fields change between runs, a stored checkpoint
+// is no longer valid because it tracks pagination state for a different query.
+//
+// Included fields: URL, and the full pagination config (mode, field names, initial_timestamp, etc.).
+// Excluded fields: auth credentials (same query, different creds), poll intervals (timing only),
+// headers, storage ID, response format/field, and metrics config.
+func configFingerprint(cfg *Config) string {
+	// Marshal the pagination config to get a stable representation of all its fields.
+	// This automatically captures any new fields added to PaginationConfig in the future.
+	paginationBytes, err := jsoniter.Marshal(cfg.Pagination)
```
eKuG: Does this mean changing page_size (e.g.,from 100 to 200) would invalidate the checkpoint and force a full re-fetch?

eKuG: `jsoniter.Marshal` on a struct with a stable field order should be deterministic in practice, but this is an implicit contract. If someone adds a `map[string]any` field to `PaginationConfig` in the future, map iteration order is non-deterministic in Go and the fingerprint would randomly change.

eKuG (changes requested):

eKuG:
Very small clarifying questions on `configFingerprint`

receiver/restapireceiver/receiver.go
Caleb-Hurshman: 1. Currently yes, that would be treated as a new query. I'll already have to update the logic once the timestamp parameters are changed by my other PR, so I can also make the fields more granular at that point.

2. This is true. What is your suggestion? I can add a comment to the pagination config struct, or were you imagining something more concrete?

eKuG: I guess adding a comment would be fine for now.

eKuG (approved):

PR #177, opened 2026-04-06
feat: Expand OCSF `int` coercion

dyl10s:
### Proposed Change

Windows events contain a lot of hex strings for numbers. Previously, these would not be converted correctly. This change adds support for the type coercion to convert hex strings (see tests).

##### Checklist
- [x] Changes are tested
- [x] CI has passed


justinianvoss22 (approved):
Looks good

PR #182, opened 2026-04-21
feat(opampexporter): add new opamp exporter

andykellr:
## Summary

- Adds `exporter/opampexporter/`, a new exporter that sends OTLP-proto payloads (logs, metrics, traces) over an existing OpAMP connection using custom messages — capability `com.bindplane.opampexporter`, message type `otlp`. This lets agents behind the OpAMP Gateway ship telemetry without a separate HTTP connection.
- Registers the custom capability once per exporter component ID; a single shared instance is reused across logs/metrics/traces pipelines. Handles `ErrCustomMessagePending` the same way `processor/snapshotprocessor` does.
- Adds a dependabot entry for the new module.

Linear: [BPOP-4973](https://linear.app/bindplane/issue/BPOP-4973/add-opampexporter-component)

## Test plan

- [x] `go build ./...` in the new module
- [x] `go vet ./...`
- [x] `go test -race ./...` — 13 tests pass (config, factory, and exporter tests covering all three signals, shared-instance registration, missing/invalid extension, pending retry, send error)
- [x] `make check-mod-paths`, `make check-dependabot`, `make check-license`


exporter/opampexporter/exporter.go
```diff
@@ -0,0 +1,159 @@
...
+	if !ok {
+		return fmt.Errorf("opamp extension %q does not exist", e.opampExtensionID)
+	}
+
+	registry, ok := ext.(opampcustommessages.CustomCapabilityRegistry)
+	if !ok {
+		return fmt.Errorf("extension %q is not a custom message registry", e.opampExtensionID)
+	}
+
+	handler, err := registry.Register(e.capability)
```
dpaasman00: Should we be using the `WithMaxQueuedMessages` option when registering?

exporter/opampexporter/exporter.go
```diff
@@ -0,0 +1,159 @@
...
+	unregisterExporter(e.exporterID)
+
+	e.mu.Lock()
+	handler := e.customCapabilityHandler
+	e.customCapabilityHandler = nil
+	e.mu.Unlock()
+
+	if handler != nil {
+		handler.Unregister()
+	}
```
dpaasman00, line 137: I don't know that it really matters a lot, but there is potential for in flight messages to be dropped during shutdown. But because we're shutting down I think it's acceptable.

dpaasman00 (commented):
A couple small comments, otherwise it looks good

exporter/opampexporter/exporter.go
andykellr: I made this a configurable option, still defaulting to 10.

exporter/opampexporter/exporter.go
andykellr: Good catch. I added locking to ensure we don't get stuck in a send loop while shutting down.

dpaasman00 (approved):

PR #189, opened 2026-04-22
fix(`restapireceiver`): Fix validation for offset/limit pagination fields

Caleb-Hurshman:
### Proposed Change

Fixes an issue where the REST API receiver was validating that the limit field name parameter was set whenever offset pagination was enabled.
The field is required when using a numeric offset to instruct the receiver how to increment the numeric value, but is irrelevant for tokenized offsets. This PR changes the validation to match the intended behavior, and updates documentation to be more descriptive.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/restapireceiver/config.go
```diff
@@ -578,8 +580,12 @@ func (c *Config) Validate() error {
 		if c.Pagination.OffsetLimit.OffsetFieldName == "" {
 			return fmt.Errorf("offset_field_name is required when pagination.mode is offset_limit")
 		}
-		if c.Pagination.OffsetLimit.LimitFieldName == "" {
-			return fmt.Errorf("limit_field_name is required when pagination.mode is offset_limit")
+		// limit_field_name is required for numeric offset pagination so the
+		// server's page size matches the internal offset advance. With
+		// token-based pagination (next_offset_field_name set), advance is
+		// driven by the token so limit_field_name is optional.
+		if c.Pagination.OffsetLimit.NextOffsetFieldName == "" && c.Pagination.OffsetLimit.LimitFieldName == "" {
```
eKuG, line 587: if NextOffsetFieldName set but user also sets StartingOffset, is that silently ignored? Worth a validation warning or doc note.

eKuG (approved):
No blockers. LGTM

receiver/restapireceiver/config.go
Caleb-Hurshman: No, it won't be ignored. StartingOffset is used for the first request, then the next offset value is read from NextOffsetFieldName 👍

PR #282, opened 2026-05-01
feat(asimstandardizationprocessor): direct source-to-ASIM Sentinel processor

eKuG:
### Proposed Change

Pairs with the Microsoft Sentinel ASIM destination merge in `bindplane-op-enterprise` (single `microsoftsentinel` destination type with `mode: basic | asim`, single consolidated DCR). One processor in the Sentinel pipeline, one DCR in Azure, one IAM role assignment. 

1. Adds `asimstandardizationprocessor`: maps source log fields directly to Microsoft Sentinel native ASIM table        
  columns. Sets `sentinel_stream_name` per-record so the Azure Log Analytics exporter routes to the right DCR stream.. Stashes the original body under `AdditionalFields` for lossless retention.                                                                                                                                                                                                  

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


exporter/azureloganalyticsexporter/log_grouper.go
```diff
@@ -0,0 +1,137 @@
...
+// TODO(routing): support per-record sentinel_endpoint attribute via a per-
+// endpoint client pool, then populate Endpoint from attribute precedence.
+type groupKey struct {
+	Endpoint   string
+	RuleID     string
+	StreamName string
+}
+
+// String returns the composite key string "endpoint|ruleID|streamName".
+func (k groupKey) String() string {
```
Caleb-Hurshman: This method is unused.

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -78,6 +85,46 @@ func (m *azureLogAnalyticsMarshaler) getRawField(ctx context.Context, field stri
...
+		return "", false
+	}
+	return v.Str(), true
+}
+
+// getStreamName resolves the DCR stream name for a log record using the
+// following precedence: log record attribute, resource attribute, config. An
+// empty attribute value is treated as unset and falls through to the next
+// source.
+func (m *azureLogAnalyticsMarshaler) getStreamName(logRecord plog.LogRecord, _ plog.ScopeLogs, resourceLog plog.ResourceLogs) string {
```
Caleb-Hurshman: Why keep the empty `plog.ScopeLogs` parameter around?

exporter/azureloganalyticsexporter/marshal.go
```diff
@@ -78,6 +85,46 @@ func (m *azureLogAnalyticsMarshaler) getRawField(ctx context.Context, field stri
...
+	if v, ok := lookupStringAttr(resourceLog.Resource().Attributes(), sentinelStreamNameAttribute); ok && v != "" {
+		return v
+	}
+	return m.cfg.StreamName
+}
+
+// getRuleID resolves the DCR rule ID for a log record using the following
+// precedence: log record attribute, resource attribute, config. An empty
+// attribute value is treated as unset and falls through to the next source.
+func (m *azureLogAnalyticsMarshaler) getRuleID(logRecord plog.LogRecord, _ plog.ScopeLogs, resourceLog plog.ResourceLogs) string {
```
Caleb-Hurshman: Same here, empty parameter

exporter/azureloganalyticsexporter/exporter.go
```diff
@@ -90,21 +99,64 @@ func (e *azureLogAnalyticsExporter) logsDataPusher(ctx context.Context, ld plog.
...
+	var errs []error
+	for key, groupLogs := range groups {
+		if groupLogs.LogRecordCount() == 0 {
+			continue
+		}
 
-	_, err = e.client.Upload(ctx, e.ruleID, e.streamName, azureLogAnalyticsLogs, nil)
-	if err != nil {
-		return e.classifyError(err)
+		// TODO(routing): support per-record sentinel_endpoint by maintaining a
```
Caleb-Hurshman: We should avoid leaving TODOs around the code

Caleb-Hurshman: Since we acknowledge that key.Endpoint is always cfg.Endpoint for now, we don't need all the validation in this PR

processor/asimstandardizationprocessor/config.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Config is the configuration for the ASIM standardization processor.
+type Config struct {
+	// EventMappings is the ordered list of event mappings.
+	EventMappings []EventMapping `mapstructure:"event_mappings"`
+}
+
+// IsKnownTargetTable returns true if the given string is a supported ASIM
+// target table.
+func IsKnownTargetTable(table string) bool {
```
Caleb-Hurshman, line 100: I'm not super familiar with ASIM yet, but should we also validate the required columns so that the processor doesn't pass on data that will fail to parse?

exporter/azureloganalyticsexporter/exporter.go
```diff
@@ -81,7 +81,16 @@ func (e *azureLogAnalyticsExporter) Capabilities() consumer.Capabilities {
...
+// logsDataPusher pushes log data to Azure Log Analytics.
+//
+// Records are first partitioned by (ruleID, streamName) based on the
+// sentinel_rule_id / sentinel_stream_name attributes (falling back to config
+// values) so that per-record routing can be honored. Each group is uploaded
+// independently; permanent errors short-circuit the remaining groups while
+// transient errors are collected and returned as a joined error so that the
+// exporterhelper retry pipeline still sees a transient failure and retries
+// the whole batch (Azure Log Analytics is expected to handle duplicates
+// idempotently on retry).
```
Caleb-Hurshman: Everything that I'm seeing says that Azure Log Analytics doesn't dedup automatically on ingest, so this seems problematic. Do you have any sources/testing that point to this working?

Caleb-Hurshman (changes requested):
Sorry I missed this while I was on vacation!
Only the first point in the PR description is implemented here, could you update that?

eKuG:
> Only the first point in the PR description is implemented here, could you update that?

Actually, 2 and 3 are also implemented, its inside the commits that I have done. So, OCSF->ASIM wasnt there earlier, I added the commit for this change and then we changed it to Raw logs -> ASIM directly, thus removing the need for OCSF processor. So its an intermittent step. 
How should I proceed? Should I remove it?

@Caleb-Hurshman

Caleb-Hurshman:
> Actually, 2 and 3 are also implemented, its inside the commits that I have done. So, OCSF->ASIM wasnt there earlier, I added the commit for this change and then we changed it to Raw logs -> ASIM directly, thus removing the need for OCSF processor. So its an intermittent step. How should I proceed? Should I remove it?

@eKuG I'm still confused about how those are implemented here. Those paths (`sentinelstandardizationprocessor ` and `examples/sentinel-asim/`) don't exist on main, and they aren't in this diff 🤔
What am I missing?

eKuG:
> > Actually, 2 and 3 are also implemented, its inside the commits that I have done. So, OCSF->ASIM wasnt there earlier, I added the commit for this change and then we changed it to Raw logs -> ASIM directly, thus removing the need for OCSF processor. So its an intermittent step. How should I proceed? Should I remove it?
> 
> @eKuG I'm still confused about how those are implemented here. Those paths (`sentinelstandardizationprocessor ` and `examples/sentinel-asim/`) don't exist on main, and they aren't in this diff 🤔 What am I missing?

@Caleb-Hurshman I added them in this PR in a commit and removed them in another commit so the code is commited but it would not be visible either on main or this PR because its a commit in this PR itself

Caleb-Hurshman:
> @Caleb-Hurshman I added them in this PR in a commit and removed them in another commit so the code is commited but it would not be visible either on main or this PR because its a commit in this PR itself

Then yes, I think you can remove those points from the description to avoid confusion 👍

exporter/azureloganalyticsexporter/exporter.go
eKuG: Good Catch for Azure Log Analytics not providing dedup. I'm trying to find a way if we can do it somehow but our current Sentinel Integration also leaves it as is. I'll keep you posted if I find something

processor/asimstandardizationprocessor/config.go
eKuG: Added some validation here : 
- Default-on validation drops records that would fail Azure ingest                                                    
- Type coercion prevents the InvalidTransformOutput → retry-storm → silent loss path entirely                         
- Failure surface: warn log per drop with target_table + missing_cols / column + value

processor/asimstandardizationprocessor/README.md
```diff
@@ -0,0 +1,204 @@
...
+   hex, decimal-point tolerant), as floats, or as booleans. Coercion failures
+   drop the offending field with a warn log; the record continues with the
+   remaining fields.
+2. **Mandatory column check.** Records missing any of the ASIM common
+   mandatory columns (`TimeGenerated`, `EventCount`, `EventStartTime`,
+   `EventEndTime`, `EventType`, `EventResult`, `EventProduct`, `EventVendor`,
+   `EventSchema`, `EventSchemaVersion`, `Dvc`) after coercion are dropped
+   entirely with a warn log.
+
+Set `runtime_validation: false` to pass mapped records through unchanged.
```
Caleb-Hurshman: I don't think this is accurate. Only the second check is changed by this field; the type-coercion check runs no matter what.
Up to you what the desired behavior is.

processor/asimstandardizationprocessor/coerce.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		for _, layout := range datetimeLayouts {
+			if t, err := time.Parse(layout, s); err == nil {
+				return t.UTC().Format(time.RFC3339Nano), true
+			}
+		}
+		return nil, false
+	case int, int32, int64, uint, uint32, uint64:
+		// Treat as unix epoch seconds.
+		sec, ok := toInt64(x)
```
Caleb-Hurshman: I'm seeing that Microsoft KQL ints are 32-bit, not 64: https://learn.microsoft.com/en-us/kusto/query/scalar-data-types/int?view=microsoft-fabric
I think this is problematic, unless ASIM is reading them automatically as `long`?

processor/asimstandardizationprocessor/coerce.go
```diff
@@ -0,0 +1,224 @@
...
+		if s == "" {
+			return nil, false
+		}
+		for _, layout := range datetimeLayouts {
+			if t, err := time.Parse(layout, s); err == nil {
+				return t.UTC().Format(time.RFC3339Nano), true
+			}
+		}
+		return nil, false
+	case int, int32, int64, uint, uint32, uint64:
```
Caleb-Hurshman: I think we need more logic to handle epoch ns, µs, and ms. Right now they could be misinterpreted as seconds and silently passed on.

processor/asimstandardizationprocessor/coerce.go
```diff
@@ -0,0 +1,224 @@
...
+func coerceString(v any) (string, bool) {
+	switch x := v.(type) {
+	case string:
+		return x, true
+	case fmt.Stringer:
+		return x.String(), true
+	case bool, int, int32, int64, uint, uint32, uint64, float32, float64:
+		return fmt.Sprint(x), true
+	default:
+		return fmt.Sprint(x), true
```
Caleb-Hurshman: Not sure what to make of this, Claude flagged:
```
A []any{1,2,3} mapped to a string column becomes "[1 2 3]" (Go's default formatting), and a map[string]any becomes "map[a:1 b:2]" — neither of which is meaningful in Sentinel and neither of which gets caught by the           mandatory-column check. The README sells this as "fmt.Sprint of any value", but in practice you almost never want to coerce a composite to a string column. Either JSON-marshal slices/maps for string columns, or fail coercion (let the warn log + field-drop catch it).
```

processor/asimstandardizationprocessor/coerce.go
```diff
@@ -0,0 +1,224 @@
...
+	default:
+		return value, true
+	}
+}
+
+func coerceString(v any) (string, bool) {
+	switch x := v.(type) {
+	case string:
+		return x, true
+	case fmt.Stringer:
```
Caleb-Hurshman: If a string column contains a time.Time, the output will be Go's debug format instead of the expected ISO timestamp. We should put this check lower and include more specific checks for types that satisfy Stringer but may transform the data in unexpected ways (time.Time, net.IP, etc.).

eKuG: Resolved

Caleb-Hurshman (approved):
Thanks for iterating. There are a couple of minor things to flag that are fine as potential follow-ups:
 - Per-record schema lookup (processor.go): asimColumnTypes[eventMapping.targetTable] inside the per-record loop. Hoist onto compiledEventMapping.                                                                                                                                    
  - Generator script reference (schema.go): comment says "regenerate via the script in this commit, " but no script is in the diff. I'd say include it or adjust the comment.                                                                                                            
  - Throwaway marshaler in log_grouper.go: &azureLogAnalyticsMarshaler{cfg: cfg} solely to reach getRuleID/getStreamName. Pull those out as package-level functions taking *Config.                                                                                                         
  - coerceReal uint64 path wraps via int64 (edge case — JSON-decoded numbers arrive as float64).                                                                                                                                                                                            
  - coerceBool doesn't handle float64 (symmetry with int cases; matters for JSON-decoded 1/0).

PR #304, opened 2026-05-08
chore(chronicleexporter): rename host metrics -> metrics

colelaven:
### Proposed Change
Renames the hostMetricsReporter in the `chronicleexporter` to just `metricsReporter` as we are now collecting more than just host metrics (eg. dropped & sent spans).

##### Checklist
- [x] Changes are tested
- [x] CI has passed


dyl10s (approved):
💯

PR #308, opened 2026-05-13
feat(lookupprocessor): add Redis, API sources and TTL cache

eKuG:
## Summary
- Extends `processor/lookupprocessor` beyond CSV with two new sources: Redis (HGETALL with JSON GET fallback, TLS, key prefix) and HTTP API (URL templating, header support, response field mapping, retry-with-backoff).
- Adds an optional TTL cache wrapping any source. Backed by an OTel storage extension (e.g. `file_storage`, `redis_storage`) when one is configured, or a per-instance in-memory map otherwise. Storage clients are namespaced per signal kind so logs/traces/metrics pipelines sharing one processor ID do not share or close each other's state. Expired entries are deleted on read for both backends.
- Ports the work originally on the legacy `danjordan/lookupprocenhancement` branch in `bindplane-otel-collector` to the segregated contrib repo, with the WRONGTYPE Redis fallback bug fixed and full unit-test coverage added (httptest for API, miniredis for Redis

## Test plan
- [x] `go build ./...` clean
- [x] `go test -race ./...` clean
- [x] `make check-fmt`, `make check-license`, `make check-mod-paths`, `make check-dependabot`, `make misspell`, `make gosec`, `revive` all clean
- [ ] Reviewer: smoke-test in collector via `make build-collector` with this branch as `COLLECTOR_PATH`'s contrib source
- [ ] Reviewer: verify Redis source against a real Redis instance with both hash and JSON value layouts
- [ ] Reviewer: verify API source against a real HTTP endpoint with `response_mapping`
- [ ] Follow-up PR on `bindplane-otel-collector` will bump `processor/lookupprocessor` to the new tag (`v1.6.0`) once this merges and is released

## Notes
- `bbolt` is not used directly; persistence is delegated entirely to OTel storage extensions.


processor/lookupprocessor/api.go
```diff
@@ -0,0 +1,226 @@
...
+		req.Header.Set(key, value)
+	}
+
+	resp, err := a.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute request: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	body, err := io.ReadAll(resp.Body)
```
Caleb-Hurshman: I think we should have a safeguard around how much data we read in here to protect from misbehaving or hostile endpoints.
`io.LimitReader()`

processor/lookupprocessor/api.go
```diff
@@ -0,0 +1,226 @@
...
+	delay := apiInitialDelay
+
+	for attempt := 0; attempt < apiMaxRetries; attempt++ {
+		if attempt > 0 {
+			a.logger.Debug("retrying API request", zap.Int("attempt", attempt), zap.Duration("delay", delay))
+			time.Sleep(delay)
+			delay *= apiRetryMultiplier
+		}
+
+		data, err := a.makeRequest(requestURL)
```
Caleb-Hurshman: All errors are treated and retried the same. We could probably save requests if we treated some errors as non-retryable (e.g., 404, 401, 400).

processor/lookupprocessor/processor.go
```diff
@@ -262,7 +308,6 @@ func (p *lookupProcessor) processGaugeMetrics(metrics pmetric.Metric) {
 	}
 }
 
-// processSummaryMetrics processes incoming summary metrics
```
Caleb-Hurshman, line 265: Any reason to remove these comments? It makes the diff quite a bit longer

processor/lookupprocessor/api.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil, fmt.Errorf("invalid URL after substitution: %w", err)
+	}
+
+	var lastErr error
+	delay := apiInitialDelay
+
+	for attempt := 0; attempt < apiMaxRetries; attempt++ {
+		if attempt > 0 {
+			a.logger.Debug("retrying API request", zap.Int("attempt", attempt), zap.Duration("delay", delay))
+			time.Sleep(delay)
```
Caleb-Hurshman: From what I can tell, this blocks the pipeline and doesn't have context awareness.
Claude recommended threading ctx through the LookupSource interface and using 
```
select { 
case <-ctx.Done(): ...
case <-time.After(delay): ...
}
```

processor/lookupprocessor/redis.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if cfg.TLS {
+		opts.TLSConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+	}
+
+	client := redis.NewClient(opts)
+
+	if err := client.Ping(context.Background()).Err(); err != nil {
```
Caleb-Hurshman: This could use a timeout to avoid blocking when DNS is slow, for example.

processor/lookupprocessor/README.md
```diff
@@ -1,25 +1,70 @@
...
-1. This processor will load and periodically refresh a CSV file in memory.
-2. When telemetry is received, the processor checks if the configured `field` exists in the configured `context`.
-3. If the field exists and the CSV contains a matching value, all other values in that record are added to the `context` of the telemetry. The header of each value is used when adding these fields.
+1. The processor loads a lookup source. Exactly one of `csv`, `redis`, or `api`
+   must be configured.
+2. When telemetry is received, the processor checks if the configured `field`
+   exists in the configured `context`.
+3. If the field exists and the source returns a match, all other key/value
+   pairs from the source row are added to the `context` of the telemetry.
+4. An optional cache (enabled by default with a 5-minute TTL) stores recent
```
Caleb-Hurshman, line 18: The 5-minute TTL is accurate, but kind of misleading. Correct me if I'm wrong, but no background sweeper cleans up entries after 5 minutes; they are only removed if that same key gets called again. So if a key is never called again, does it live in memory forever?

Caleb-Hurshman (changes requested):
1. Could you link to a linear ticket/project? I don't recall the context for this change super clearly.
2. Let's sync on how to go over some of the reviewer testing with Redis, etc.

eKuG:
> 1. Could you link to a linear ticket/project? I don't recall the context for this change super clearly.
> 2. Let's sync on how to go over some of the reviewer testing with Redis, etc.

Way to test this : 

1. Wire repos: in contrib, echo "COLLECTOR_PATH=../bindplane-otel-collector" > .local.env.
2. Seed Redis:
    - docker run -d --name lookup-redis -p 6379:6379 redis:7-alpine
    - hash form: docker exec lookup-redis redis-cli HSET user:42 name alice team sre region us-west
    - string form: docker exec lookup-redis redis-cli SET 'ip:1.2.3.4' '{"host":"host-a","env":"prod"}' 
3. Build collector (workspace overrides v1.5.0 pin to local contrib source):
    - in contrib: make _build-setup
    - go build -tags bindplane -o /tmp/lookup-collector ~/Documents/bindplane-otel-collector/cmd/collector
4. Input log at /tmp/lookup-test/input.log — three JSON lines: one w/ user_id:42, one w/ user_id:99 (missing key), one
   w/ ip:1.2.3.4.
5. Config at /tmp/lookup-test/config.yaml: filelog → lookup (redis, key_prefix per source) → debug exporter. Must set
  service.telemetry.metrics.level: none to avoid 8888 bind clash.
6. Run + observe: /tmp/lookup-collector --config=... > /tmp/lookup.log 2>&1 &, sleep 5, kill. grep -A1 "ResourceLog #"
   /tmp/lookup.log — verify enriched Attributes: block.

processor/lookupprocessor/api_test.go
```diff
@@ -0,0 +1,215 @@
...
+	start := time.Now()
+	_, err = src.Lookup(ctx, "any")
+	elapsed := time.Since(start)
+	require.Error(t, err)
+	// Without ctx, retries would take 100ms+200ms=300ms. Cancel during the first backoff
+	// must return before the second attempt completes its full delay budget.
+	require.Less(t, elapsed, 250*time.Millisecond, "ctx cancel must abort pending retry sleep")
+}
+
+func TestAPISourceLookup_BodyTruncatedAtLimit(t *testing.T) {
```
Caleb-Hurshman, line 176: This isn't exactly testing the truncation anymore. This test could be removed, or enhanced to test a scenario where the handler streams more data than the limit reader allows

processor/lookupprocessor/redis.go
```diff
@@ -0,0 +1,149 @@
...
+			MinVersion: tls.VersionTLS12,
+		}
+	}
+
+	client := redis.NewClient(opts)
+
+	pingCtx, cancel := context.WithTimeout(context.Background(), redisBootPingTimeout)
+	defer cancel()
+	if err := client.Ping(pingCtx).Err(); err != nil {
+		logger.Warn("redis ping failed at start; source will retry per lookup, cached entries will still be served",
```
Caleb-Hurshman: I feel like this should fail fast so that a misconfigured redis connection can't start.
It's good that the cache can still work, but lookups wouldn't work, and it could take a while for a customer to realize (and then look at the logs).

processor/lookupprocessor/api.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const (
+	apiMaxRetries      = 3
+	apiInitialDelay    = 100 * time.Millisecond
+	apiRetryMultiplier = 2
+
+	// apiMaxResponseBytes caps how much of a response body the source will read
+	// to protect against misbehaving or hostile endpoints returning huge payloads.
+	apiMaxResponseBytes = 1 << 20 // 1 MiB
```
Caleb-Hurshman, line 40: Can we make this configurable?

processor/lookupprocessor/redis.go
```diff
@@ -0,0 +1,149 @@
...
+	"strings"
+	"time"
+
+	"github.com/redis/go-redis/v9"
+	"go.uber.org/zap"
+)
+
+// redisLookupTimeout bounds each Redis call so a slow or unreachable server
+// cannot stall pipeline processing indefinitely.
+const redisLookupTimeout = 5 * time.Second
```
Caleb-Hurshman: Another thing that's normally configurable

processor/lookupprocessor/api.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+// nonRetryableStatusError signals an HTTP status that should not be retried.
+type nonRetryableStatusError struct {
+	status int
+	body   string
+}
+
+func (e *nonRetryableStatusError) Error() string {
```
Caleb-Hurshman, line 54: Claude flagged:
```
nonRetryableStatusError stores up to 1 MiB and embeds it in Error(). A burst of failing lookups will allocate megabytes for error strings and
  write them to logs. Truncate the captured body to ~256 bytes for the error message.
```

Not sure how common that is, but it seems like a good safeguard

processor/lookupprocessor/README.md
```diff
@@ -1,25 +1,77 @@
...
+   lookups. The cache backend is either an OpenTelemetry storage extension
+   (e.g. `file_storage`, `redis_storage`) for persistence across restarts, or
+   a per-instance in-memory map when no `storage` is configured.
+
+   Eviction is lazy: there is no background sweeper. An expired entry is
```
Caleb-Hurshman, line 23: Good addition to the docs 👍

processor/lookupprocessor/processor.go
```diff
@@ -227,70 +283,69 @@ func (p *lookupProcessor) processMetricsWithAttributesContext(md pmetric.Metrics
...
 		attrs := exponentialHistogram.DataPoints().At(i).Attributes()
-		p.addLookupValues(attrs)
+		p.addLookupValues(ctx, attrs)
 	}
 }
 
-// addLookupValues adds lookup values to the source map
-func (p *lookupProcessor) addLookupValues(source pcommon.Map) {
+// addLookupValues adds lookup values to the source map.
+func (p *lookupProcessor) addLookupValues(ctx context.Context, source pcommon.Map) {
```
Caleb-Hurshman, line 348: If the passed-in context has no deadline, the API could stall for up to 30s (3 retries * 10s each).
The redis source has a 5s bound per lookup, it would be good to have the same thing for the API

Caleb-Hurshman (changes requested):

Caleb-Hurshman (approved):
Nice work ✅

PR #320, opened 2026-05-21
fix(chronicleexporter): merge attribute & config ingestion labels in http exporter

colelaven:
### Proposed Change
Previously, the logic when aggregating ingestion labels in the chronicle exporter behaved differently for https protocol compared to grpc protocol. For gRPC, the ingestion labels present in the config parameters were merged with those present in a map at the special field `attributes['chronicle_ingestion_label']` on a log. However for HTTPS, if any labels were present in the special attribute field, only those were used when imports logs to the chronicle API, and the labels defined in the config were ignored.

This PR:
- Aligns ingestion label aggregation logic for HTTPS to be the same as gRPC: Merges the ingestion labels present in the special attribute field with those defined in the exporter config, with a precedence for those from the attribute.
- Updates the Readme to better describe this behavior.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/exporter/chronicleexporter/README.md
+++ b/exporter/chronicleexporter/README.md
@@ -72,7 +72,7 @@ If the `attributes["chronicle_log_type"]` field is present in the log, its value
 
 If the `attributes["chronicle_namespace"]` field is present in the log, its value will be used in the payload instead of the `namespace` in the config.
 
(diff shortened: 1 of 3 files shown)


shazlehu (approved):
Works as expected: attribute `chronicle_ingestion_label["key1"]` takes precedence over advanced ingestion labels in the config, config values present when attribute is absent.

PR #413, opened 2026-06-09
feat(awsneuronreceiver): add AWS Neuron metrics receiver (neuron-monitor + sysfs)

Dylan-M:
Implements the AWS Neuron receiver for project [AWS Neuron Metrics Source](https://linear.app/bindplane/project/aws-neuron-metrics-source-d1874119bf17): PIPE-1077, PIPE-1078, PIPE-1079, PIPE-1080, and docs (PIPE-1082). The Bindplane source-type (PIPE-1084) ships separately.

### Proposed Change
Adds `receiver/awsneuronreceiver`, a metrics receiver for AWS Neuron
accelerators (Inferentia/Trainium), with two collection streams that emit
distinct metrics:

- **Primary (neuron-monitor):** spawns the vendor `neuron-monitor` binary and
  parses its JSON for NeuronCore utilization, FLOPS, execution latency
  percentiles, execution counts/errors, per-core & runtime memory, ECC, and vCPU.
- **Secondary (sysfs):** pure-Go, read-only reader for the finer-grained
  11-category device-memory breakdown (present/peak) the monitor doesn't expose,
  plus ECC/topology as a fallback when the binary isn't installed.

All metrics use the `aws.neuron.*` namespace and follow the OTel semantic
conventions (utilization as a ratio unit `1`, bytes `By`, durations `s`, counts
as `{...}` monotonic sums, `error.type`/`hw.id` from the hardware conventions).
The full catalog is defined in `metadata.yaml` (mdatagen) with a curated
default-on subset; a two-layer config (`metric_groups` category toggles +
per-metric overrides) gates the rest.

The receiver degrades gracefully, and each stream fails independently: if
`neuron-monitor` is absent, fails to start, or its stream ends, it logs once and
continues on sysfs; if the sysfs tree is unreadable, it logs once and continues on
`neuron-monitor` (individual missing sysfs files are logged at debug and skipped).
A single-path failure never brings down the collector; if **both** paths fail, the receiver surfaces a scrape error each interval (like other scraper receivers) rather than emitting nothing. See the README.

Validated end-to-end on a real inf2.xlarge under a torch-neuronx workload.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### Test artifacts
Snapshot BDOT builds with this receiver compiled in (receiver commit `c73139cb`), hardware-tested on an inf2.xlarge under load: the full runtime metric set emits with correct resource attributes, and the unified `collection_interval` → neuron-monitor `period` is confirmed on hardware. The changes since that validation are config, metadata, and error-handling refinements — the `aws.neuron.device.power.utilization` min/max/avg split (`aws.neuron.power.statistic` attribute), the inherited 60s default `collection_interval`, the receiver owning the neuron-monitor config (no external `config_file`), and a scrape error when both collection paths fail — none of which alter runtime metric emission, and all covered by unit tests.

<details>
<summary>Download links (linux: amd64, arm64, arm, ppc64, ppc64le)</summary>

**amd64**
- [tar.gz](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector-v1.101.1-SNAPSHOT-84477ef0c-linux-amd64.tar.gz) | [deb](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_amd64.deb) | [rpm](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_amd64.rpm)

**arm64**
- [tar.gz](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector-v1.101.1-SNAPSHOT-84477ef0c-linux-arm64.tar.gz) | [deb](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_arm64.deb) | [rpm](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_arm64.rpm)

**arm**
- [tar.gz](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector-v1.101.1-SNAPSHOT-84477ef0c-linux-arm.tar.gz) | [deb](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_arm.deb) | [rpm](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_arm.rpm)

**ppc64**
- [tar.gz](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector-v1.101.1-SNAPSHOT-84477ef0c-linux-ppc64.tar.gz) | [deb](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_ppc64.deb) | [rpm](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_ppc64.rpm)

**ppc64le**
- [tar.gz](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector-v1.101.1-SNAPSHOT-84477ef0c-linux-ppc64le.tar.gz) | [deb](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_ppc64le.deb) | [rpm](https://storage.googleapis.com/neuron-receiver/v1.101.1-SNAPSHOT-84477ef0c-receiver-c73139cb/observiq-otel-collector_v1.101.1-SNAPSHOT-84477ef0c_linux_ppc64le.rpm)

</details>

These links are refreshed whenever the artifacts are rebuilt; the receiver commit in the path identifies the build.


receiver/awsneuronreceiver/metadata.yaml
```diff
@@ -0,0 +1,271 @@
...
+    attributes: [error_type]
+  aws.neuron.execution.latency:
+    enabled: true
+    stability: alpha
+    description: Execution latency percentiles reported by neuron-monitor.
+    unit: s
+    gauge:
+      value_type: double
+    attributes: [latency_type, quantile]
+  aws.neuron.monitor.memory.usage:
```
eKuG: metric defined but NEVER recorded Catalogued in metadata.yaml and documentation.md but I see no `RecordAwsNeuronMonitorMemoryUsage*` call exists in scraper.go / sysfs.go / neuronmonitor.go

receiver/awsneuronreceiver/neuronmonitor.go
```diff
@@ -0,0 +1,223 @@
...
+	"io"
+	"os/exec"
+	"sync"
+	"sync/atomic"
+
+	"go.uber.org/zap"
+)
+
+// maxLineBytes bounds a single neuron-monitor JSON report line. The default
+// bufio.Scanner token cap (64 KiB) can be exceeded when many models are loaded.
```
eKuG: bufio.Scanner line-based JSON parse is fragile
The implementation assumes each report is exactly one line. AWS neuron-monitor default output is newline-separated single-line JSON, so the default config works but there are two failure modes:

Pretty-printed configs break silently. Multi-line JSON per report causes the scanner to split at internal newlines, so json.Unmarshal fails and reports are lost forever. This is logged at Debug only, not Error, so the loss is invisible in normal operation.
maxLineBytes = 4 MiB caps report size. Multi-model hosts can produce reports that exceed this limit.

Fix: use json.NewDecoder(stdout).Decode(&rep) in a loop instead. It handles arbitrary whitespace, has no size cap, and is robust to pretty-printing.

receiver/awsneuronreceiver/scraper.go
```diff
@@ -0,0 +1,154 @@
...
+	if mi.MemoryTotalBytes > 0 {
+		s.mb.RecordAwsNeuronSystemMemoryUsageDataPoint(now, mi.MemoryUsedBytes, "used")
+		s.mb.RecordAwsNeuronSystemMemoryUsageDataPoint(now, mi.MemoryTotalBytes, "total")
+	}
+	for state, v := range rep.SystemData.VcpuUsage.AverageUsage {
+		s.mb.RecordAwsNeuronSystemCPUUtilizationDataPoint(now, v/100.0, "average", state)
+	}
+}
+
+func setResourceFromReport(rb *metadata.ResourceBuilder, rep *nmReport) {
```
eKuG, line 156: Receiver overrides cloud.* / host.* resource attributes from neuron-monitor
setResourceFromReport unconditionally sets cloud.provider=aws, host.id, host.type, cloud.region, and cloud.availability_zone. This conflicts with the universally-recommended resourcedetectionprocessor that most production deployments run.
Stamping these attributes in the receiver means two sources are writing the same keys, last write wins, and the outcome depends on pipeline order, which is fragile.

Fix, pick one:
1. Gate the behavior behind a set_resource_attributes config option (default false), or
2. Drop it entirely and let resourcedetection handle these attributes.

Either way, document the choice.

receiver/awsneuronreceiver/scraper.go
```diff
@@ -0,0 +1,154 @@
...
+	if hi := rep.NeuronHardwareInfo; hi.NeuronDeviceType != "" {
+		rb.SetAwsNeuronDeviceType(hi.NeuronDeviceType)
+		rb.SetAwsNeuronNeuroncoreVersion(hi.NeuroncoreVersion)
+	}
+}
+
+func coreToInt(coreKey string) int64 {
+	n, _ := strconv.ParseInt(coreKey, 10, 64)
+	return n
+}
```
eKuG: swallows ParseInt error. Returns 0 on parse failure. Non-numeric core key collapses all such entries to core=0, masking real data.

receiver/awsneuronreceiver/neuronmonitor.go
```diff
@@ -0,0 +1,223 @@
...
+		cancel()
+		return
+	}
+
+	r.wg.Add(1)
+	go func() {
+		defer r.wg.Done()
+		r.consume(ctx, stdout)
+		// Reap the process so we never leave a zombie.
+		_ = cmd.Wait()
```
eKuG: The code calls `_ = cmd.Wait()` after `consume` returns, so the exit code is lost. A non-zero exit from the `neuron-monitor` process,bad `config_file` path, missing permissions is the typical "binary won't start properly" case, and the wait error carries the exit code plus stderr.
Fix: capture stderr via `cmd.StderrPipe()` and log any non-zero exit at Error level with a tail of stderr. Without this, field debugging is impossible.

eKuG (changes requested):

receiver/awsneuronreceiver/metadata.yaml
Dylan-M: Removed it. There's no real data source for this in neuron-monitor's output (no `self_stats` block appears in the captured reports), so rather than wire a metric that can never be recorded, I dropped it from metadata.yaml and regenerated. Fixed in 848e6b9f.

receiver/awsneuronreceiver/neuronmonitor.go
Dylan-M: Done. Switched to a `json.NewDecoder(stdout).Decode()` loop, so pretty-printed reports parse correctly and there's no per-report size cap. Added `TestConsumeParsesPrettyPrintedReports`. Fixed in 848e6b9f.

receiver/awsneuronreceiver/scraper.go
Dylan-M: Good catch on the others. On the resource attributes, I don't think this is a true conflict. `resourcedetectionprocessor` has an `override` option (default `true`) that deterministically resolves the overlap, so it isn't last-write-wins by pipeline order. With the default, detection wins; with `override: false`, the receiver's values are kept. The receiver stamps `cloud.*`/`host.*` from the IMDS data neuron-monitor already reports, so the metadata stays correct even when no detection processor is configured (which we can't assume). I've documented the interaction in the README rather than gating it behind a config flag. Happy to revisit if you see a case the `override` semantics don't cover.

receiver/awsneuronreceiver/scraper.go
Dylan-M: Fixed. `coreToInt` now returns `(int64, bool)`, and callers skip plus Debug-log a non-numeric key instead of collapsing it onto core 0. Added `TestRecordMonitorSkipsNonNumericCoreKey`. Fixed in 848e6b9f.

receiver/awsneuronreceiver/neuronmonitor.go
Dylan-M: Fixed. stderr is now captured via a bounded tail buffer set as `cmd.Stderr`, and a non-zero exit is logged once at Error with the stderr tail. Added `TestRunnerLogsStderrOnNonZeroExit`. Fixed in 848e6b9f.

eKuG (dismissed):
LGTM

eKuG (approved):
LGTM

Dylan-M:
### Merge activity

* **Jun 15, 5:49 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/413).

PR #458, opened 2026-06-16
feat(awss3eventreceiver): auto-detect gzip by magic number

eKuG:
### Proposed Change

The AWS S3 Event source only decompressed objects flagged by a `Content-Encoding: gzip` header or a `.gz` key suffix. Producers such as the **AWS Landing Zone Accelerator (LZA)** export gzip-compressed CloudWatch logs to S3 with neither, so the compressed bytes flowed downstream (e.g. to Google SecOps) as an unparseable log type, causing missing detections. Affected customers cannot change the LZA export format.

`LogStream.BufferedReader` now adds content-based detection as a third step. Detection order:

1. `Content-Encoding: gzip` header
2. `.gz` object key suffix
3. **new** — gzip magic number (`0x1f 0x8b 0x08`) at the start of the object body, used only when neither of the above is present

Detection is automatic with no new configuration. Existing header/extension handling is unchanged, and non-compressed objects are passed through untouched (the three-byte magic, the two ID bytes plus the deflate method, makes a false positive on real log data effectively impossible).

**Restart/resume:** detection always runs on the raw object at byte 0, the worker always issues a full `GetObject` (no `Range` header), and decompression precedes the resume offset-skip while the offset is counted in decompressed bytes. First read and resume behave identically.

Scope is limited to this receiver. No source/UI changes are required in bindplane-op-enterprise.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/awss3eventreceiver/internal/worker/stream.go
```diff
@@ -35,32 +43,50 @@ type LogStream struct {
...
+	}
+	return NewBufferedReader(br, stream.MaxLogSize), nil
+}
+
+// gzipReader wraps r in a gzip decompressor and returns a BufferedReader over
+// the decompressed stream.
+func (stream *LogStream) gzipReader(r io.Reader) (BufferedReader, error) {
+	gz, err := gzip.NewReader(r)
+	if err != nil {
+		return nil, fmt.Errorf("create gzip reader: %w", err)
```
Dylan-M, line 89: Coverage on the new code: `BufferedReader` is 100%. `gzipReader` is 75%, with the `gzip.NewReader` error return (stream.go:88-90) untested. It's reachable when a body carries a gzip signal (the header or a `.gz` suffix) but isn't valid gzip. Could you add a case with `ContentEncoding: gzip` over a non-gzip body and assert the error?

Dylan-M (changes requested):

receiver/awss3eventreceiver/internal/worker/stream.go
eKuG: Addressed

Dylan-M (approved):
Error path is covered now and the gzip reader reads 100%. Thanks for adding the cases.

PR #460, opened 2026-06-16
feat(chronicleexporter): expose HTTP response header timeout

colelaven:
### Proposed Change
- Exposes the transport level `ResponseHeaderTimeout` field in the Chronicle Exporter for advanced configuration. This only applies to the HTTPS protocol (Chronicle API) variation of the Chronicle Exporter.

This value may need to be increased from the default of 10 seconds in high-latency environments to prevent `http2: timeout awaiting response headers` errors from occurring when making requests to the Chronicle API.

##### Checklist
- [x] Changes are tested
- [x] CI has passed


jsirianni (approved):
Thanks!

PR #478, opened 2026-06-25
fix(m365receiver): tolerate auth failures instead of failing collector startup

schmikei:
### Proposed Change
A component's start should not rely on external factors like token expiry or anything on that front otherwise the collector will fail to start with what is considered a valid configuration at one moment but invalid at a later date. 

This changes the m365 receiver component to not have this state and effectively logs error messages rather than failing to start.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/m365receiver/logs.go
+++ b/receiver/m365receiver/logs.go
@@ -150,7 +150,7 @@ func (l *m365LogsReceiver) ensureToken(ctx context.Context) error {
 
 	for _, a := range l.audits {
 		if err := l.client.StartSubscription(ctx, l.startRoot+a.route); err != nil {
-			return fmt.Errorf("error starting %s audit subscription: %w", a.name, err)
+			return fmt.Errorf("starting %s audit subscription: %w", a.name, err)
 		}
 	}
 
--- a/receiver/m365receiver/logs.go
+++ b/receiver/m365receiver/logs.go
@@ -61,6 +61,12 @@ type m365LogsReceiver struct {
 	record       *logRecord
 	root         string
 	startRoot    string
+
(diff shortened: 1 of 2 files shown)


receiver/m365receiver/logs.go
```diff
@@ -131,9 +125,39 @@ func (l *m365LogsReceiver) Start(ctx context.Context, host component.Host) error
...
+// on the M365 side, so a partially-completed attempt is retried in full on the
+// next call. On success it marks the receiver connected and becomes a no-op
+// thereafter.
+func (l *m365LogsReceiver) ensureToken(ctx context.Context) error {
+	if l.connected {
+		return nil
+	}
+
+	if err := l.client.GetToken(ctx); err != nil {
+		return fmt.Errorf("error creating authorization token: %w", err)
```
dpaasman00:
```suggestion
		return fmt.Errorf("creating authorization token: %w", err)
```

receiver/m365receiver/logs.go
```diff
@@ -131,9 +125,39 @@ func (l *m365LogsReceiver) Start(ctx context.Context, host component.Host) error
...
+	}
+
+	for _, a := range l.audits {
+		if err := l.client.StartSubscription(ctx, l.startRoot+a.route); err != nil {
+			return fmt.Errorf("error starting %s audit subscription: %w", a.name, err)
```
dpaasman00:
```suggestion
			return fmt.Errorf("starting %s audit subscription: %w", a.name, err)
```

dpaasman00 (approved):

PR #555, opened 2026-07-20
chore: Add weekly govulncheck workflow

dpaasman00:
### Proposed Change

Duplicate the govulncheck workflow from bindplane-otel-collector, scanning every module in the repo and posting results to Slack.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


.github/workflows/govulncheck.yaml
```diff
@@ -0,0 +1,110 @@
...
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      - name: Setup Go
+        uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version-file: "internal/tools/go.mod"
+          cache-dependency-path: |
+            **/go.sum
+
+      - name: Get official govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@v1.1.4
```
mrsillydog: Latest appears to be 1.6.0. Can we consider how we're going to keep this reference up-to-date?

dpaasman00: I think I'll just add it to our internal/tools list and then run 'make install-tools' before this action runs. Then dependabot will handle keeping it up to date

mrsillydog (approved):

PR #628, opened 2026-07-27
ci: make build/test summary gates tolerant of cancelled runs (BPOP-4734)

Dylan-M:
### Proposed Change

This is a preventative change. It has no effect on current CI. If we ever add `concurrency: cancel-in-progress` (which I think we should to save CI runs), this will prevent the same false notification spam that was experienced on the BP repo.

##### Checklist
- [ ] Changes are tested (N/A)
- [x] CI has passed

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -116,13 +116,41 @@ jobs:
     if: always()
     needs: [optimize_ci, build-collector]
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
     steps:
       - name: Check build results
+        env:
+          GH_TOKEN: ${{ github.token }}
+          CURRENT_RUN_ID: ${{ github.run_id }}
+          BRANCH: ${{ github.head_ref || github.ref_name }}
+          REPO: ${{ github.repository }}
+          RESULT: ${{ needs.build-collector.result }}
         run: |
-          if [[ "${{ needs.build-collector.result }}" == "success" ]]; then
+          if [[ "$RESULT" == "success" ]]; then
             echo "All builds succeeded"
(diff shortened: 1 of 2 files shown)


Dylan-M:
* **#628** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/628?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/628?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/build-check.yml
```diff
@@ -116,13 +116,41 @@ jobs:
     if: always()
```
Caleb-Hurshman, line 116: Replacing if: always() with if: ${{ !cancelled() }} on the summary job skips the gate on cancelled runs with no API call, no actions: read grant, and no false green. The only thing lost is "manual cancel with no newer run goes red", which the PR's heuristic already provides unreliably. Cancelled matrix check-runs remain visible either way.

.github/workflows/tests.yml
```diff
@@ -179,13 +179,41 @@ jobs:
     if: always()
```
Caleb-Hurshman, line 179: Same here

Caleb-Hurshman (changes requested):

Dylan-M:
Closing for now. We're deferring this until the BDOT Contrib merge-queue decision lands, since the right approach (this heuristic vs a simpler `!cancelled()` skip) depends on whether a required merge-queue check is coming. I'll reopen when we're ready.

PR #669, opened 2026-07-30
chore(ci): retry transient Go module fetches (PIPE-1058)

Dylan-M:
### Proposed Change

Retries Go module fetches on transient proxy/network errors, so a module-proxy blip no longer fails a job. A shared `.github/scripts/retry.sh` wraps the `make install-tools` and `install-tools-ci` steps in the checks, tests, and govulncheck workflows, plus the `build-collector` build step. The wrapper retries only on a transient-network signature, so a real build failure still fails immediately. (PIPE-1058)

The trigger was a `build-collector/windows` run that failed on a `proxy.golang.org` HTTP/2 stream error while fetching a module.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/.github/scripts/retry.sh
+++ b/.github/scripts/retry.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+# Copyright observIQ, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# retry.sh — run a command, retrying only on transient network/registry failures.
+#
+# Absorbs the CI flakes tracked in PIPE-1058 (Go module-proxy blips:
+# proxy.golang.org HTTP/2 stream errors, i/o timeouts) and PIPE-1092
+# (Docker Hub pull timeouts to registry-1.docker.io). Bounded, so a persistent
+# outage still fails the job after RETRY_MAX attempts.
+#
+# By default it retries ONLY when the command output matches a known transient
+# signature, so wrapping a build/test/install step never masks a real compile
+# or test failure. Set RETRY_ANY=1 to retry on any non-zero exit.
+#
+# Usage: bash .github/scripts/retry.sh <command> [args...]
+# Env:   RETRY_MAX (default 3), RETRY_DELAY base backoff seconds (default 10),
+#        RETRY_ANY (default 0).
+set -uo pipefail
+
+max="${RETRY_MAX:-3}"
+base_delay="${RETRY_DELAY:-10}"
+retry_any="${RETRY_ANY:-0}"
+
+# Signatures of transient network/registry failures that clear on a retry.
+transient_re='proxy\.golang\.org|sum\.golang\.org|registry-1\.docker\.io|registry\.docker\.io|i/o timeout|TLS handshake timeout|connection reset|connection refused|unexpected EOF|stream error|INTERNAL_ERROR|net/http: request canceled|Client\.Timeout exceeded|Service Unavailable|Bad Gateway|Gateway Time-?out|Too Many Requests|temporary failure|no such host'
+
+attempt=1
+while true; do
+  log="$(mktemp)"
+  "$@" 2>&1 | tee "$log"
+  status="${PIPESTATUS[0]}"
+  if [ "$status" -eq 0 ]; then
+    rm -f "$log"
+    exit 0
+  fi
+
+  if [ "$attempt" -ge "$max" ]; then
+    echo "retry: '$*' failed after ${max} attempt(s) (exit ${status}); giving up." >&2
+    rm -f "$log"
+    exit "$status"
+  fi
+
+  if [ "$retry_any" != "1" ] && ! grep -qiE "$transient_re" "$log"; then
+    echo "retry: '$*' failed (exit ${status}) with no transient-network signature; not retrying." >&2
+    rm -f "$log"
+    exit "$status"
+  fi
+  rm -f "$log"
+
+  delay=$(( base_delay * attempt ))
+  echo "retry: transient failure on attempt ${attempt}/${max} (exit ${status}); retrying '$*' in ${delay}s..." >&2
+  sleep "$delay"
+  attempt=$(( attempt + 1 ))
+done
--- a/.github/scripts/retry.sh
+++ b/.github/scripts/retry.sh
@@ -15,14 +15,12 @@
 
 # retry.sh — run a command, retrying only on transient network/registry failures.
 #
-# Absorbs the CI flakes tracked in PIPE-1058 (Go module-proxy blips:
-# proxy.golang.org HTTP/2 stream errors, i/o timeouts) and PIPE-1092
-# (Docker Hub pull timeouts to registry-1.docker.io). Bounded, so a persistent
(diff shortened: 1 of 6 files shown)


Dylan-M:
* **#669** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/scripts/retry.sh
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# retry.sh — run a command, retrying only on transient network/registry failures.
+#
+# Absorbs the CI flakes tracked in PIPE-1058 (Go module-proxy blips:
```
Caleb-Hurshman: Remove all references to the related ticket

.github/workflows/tests.yml
```diff
@@ -170,7 +170,7 @@ jobs:
             ${{ runner.os }}-go-tools-
 
       - name: Install Tools
-        run: make install-tools-ci
+        run: bash .github/scripts/retry.sh make install-tools-ci
```
Caleb-Hurshman, line 173: CLRF will break this flow on Windows.
 The repo has no .gitattributes, and GitHub-hosted Windows images ship Git with core.autocrlf=true.

.github/scripts/retry.sh
```diff
@@ -0,0 +1,66 @@
...
+# Env:   RETRY_MAX (default 3), RETRY_DELAY base backoff seconds (default 10),
+#        RETRY_ANY (default 0).
+set -uo pipefail
+
+max="${RETRY_MAX:-3}"
+base_delay="${RETRY_DELAY:-10}"
+retry_any="${RETRY_ANY:-0}"
+
+# Signatures of transient network/registry failures that clear on a retry.
+transient_re='proxy\.golang\.org|sum\.golang\.org|registry-1\.docker\.io|registry\.docker\.io|i/o timeout|TLS handshake timeout|connection reset|connection refused|unexpected EOF|stream error|INTERNAL_ERROR|net/http: request canceled|Client\.Timeout exceeded|Service Unavailable|Bad Gateway|Gateway Time-?out|Too Many Requests|temporary failure|no such host'
```
Caleb-Hurshman: This can still catch permanent errors (e.g. 404 Not Found reading https://proxy.golang.org when contrib is ahead of the collector)
Better to match the transient error class and exclude 4xx, checksum mismatch, unknown revision, etc.

Caleb-Hurshman (changes requested):

.github/scripts/retry.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Permanent failures that never clear on retry (missing-version 404/410, checksum
+# mismatch, unknown revision). Checked before transient_re and wins, so a permanent
+# error naming a transient host is not retried. 429/5xx stay transient.
+permanent_re='40[0-9] (Bad Request|Unauthorized|Forbidden|Not Found|Method Not Allowed)|410 Gone|unknown revision|invalid version|no matching versions|missing go\.sum entry|checksum mismatch|SECURITY ERROR|verifying module|manifest unknown'
```
Caleb-Hurshman: `verifying module` disables retries for all sum.golang.org flakes. That prefix is used for all sumdb errors, including transient.
Also redundant, `checksum mismatch` and `SECURITY ERROR` already cover the permanent sumdb case

Caleb-Hurshman: Same over match for `invalid version`, should be tightened to `invalid version: unknown revision`

.github/scripts/retry.sh
```diff
@@ -0,0 +1,76 @@
...
+  fi
+
+  if [ "$attempt" -ge "$max" ]; then
+    echo "retry: '$*' failed after ${max} attempt(s) (exit ${status}); giving up." >&2
+    rm -f "$log"
+    exit "$status"
+  fi
+
+  if [ "$retry_any" != "1" ]; then
+    if grep -qiE "$permanent_re" "$log"; then
```
Caleb-Hurshman: Both the retry and permanent regexes match the whole log, and permanent wins.
Problematic in cases like `build collector`, where a transient proxy failure in tidy leaves an incomplete `go.sum`, making `go build` produce a permanent `missing go.sum...` error.
Retry is refused, even though a transient error was the cause.

.github/scripts/retry.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Permanent failures that never clear on retry (missing-version 404/410, checksum
+# mismatch, unknown revision). Checked before transient_re and wins, so a permanent
+# error naming a transient host is not retried. 429/5xx stay transient.
+permanent_re='40[0-9] (Bad Request|Unauthorized|Forbidden|Not Found|Method Not Allowed)|410 Gone|unknown revision|invalid version|no matching versions|missing go\.sum entry|checksum mismatch|SECURITY ERROR|verifying module|manifest unknown'
+
+attempt=1
+while true; do
+  log="$(mktemp)"
+  "$@" 2>&1 | tee "$log"
```
Caleb-Hurshman, line 51: Nit: there's no `trap 'rm -f "$log"' EXIT`, so a cancelled job will leak the temp file

Caleb-Hurshman (changes requested):

.github/scripts/retry.sh
```diff
@@ -0,0 +1,66 @@
...
+
+max="${RETRY_MAX:-3}"
+base_delay="${RETRY_DELAY:-10}"
+retry_any="${RETRY_ANY:-0}"
+
+# Transient network/registry failures that clear on a retry. Match error
+# signatures, not bare hostnames: a permanent 404 names the same host as a
+# transient blip, so matching the host would wrongly retry it. A permanent error
+# carries no signature here, so it falls through and is not retried.
+transient_re='i/o timeout|TLS handshake timeout|connection reset|connection refused|unexpected EOF|stream error|INTERNAL_ERROR|net/http: request canceled|Client\.Timeout exceeded|Internal Server Error|Service Unavailable|Bad Gateway|Gateway Time-?out|Too Many Requests|temporary failure|no such host'
```
Caleb-Hurshman: Removing `proxy\.golang\.org` was right for the 404 problem, but nothing replaced the host-only signatures. Both now fall through as no transient-network signature; not retrying:
- Get "https://proxy.golang.org/x/@v/list": EOF — a bare EOF, arguably the most common go-proxy blip
- http2: server sent GOAWAY and closed the connection

Something like `EOF$|GOAWAY|http2:|Could not resolve host` should be an improvement over dropping hosts outright.

.github/scripts/retry.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# One temp file for the run; the trap clears it on normal exit and on cancellation.
+log="$(mktemp)"
+trap 'rm -f "$log"' INT TERM EXIT
```
Caleb-Hurshman: Regression, this swallows cancellation, confirmed by Claude:
Verified by sending SIGTERM during the backoff sleep:
RESULT: retry.sh SURVIVED SIGTERM and is still running
... retry: transient failure on attempt 2/3 ...; retrying in 10s...
... failed after 3 attempt(s)
On a cancelled Actions job or step timeout the runner signals the process group: the wrapped make dies, retry.sh reads that as a failed attempt, and if the log carries a transient signature it launches a fresh multi-platform collector build after the job was cancelled, until the hard SIGKILL. Before this commit the script died on the signal. 
Fix: re-raise — trap 'rm -f "$log"; exit 130' INT / exit 143 for TERM — or keep EXIT only.

Caleb-Hurshman (approved):

PR #675, opened 2026-08-04
chore: Add JSON fingerprinting function to log type detection processor

dyl10s:
### Proposed Change
This adds the fingerprinting function and a custom function for the first type, JSON. The goal here is to create a fingerprint that does a decent job at just grabbing the log structure. In json that is pretty simple, the special characters defining the json structure, `{` `[`, and the keys paint a good picture.

The goal of this is not to be perfect but good enough where we dont have to do log type detection for every single log. Speed and memory is very important here because this fingerprinting function will run on every log. The key added to the log wont always be there but is nice for testing right now.

This PR has the following
- JSON test logs created with Claude, 2 for each source, 250 different sources. Generated by 2 different agents to avoid bias
- Fingerprint JSON function that does a single pass walk through the log body and generates a hash based on the structure
- [FNV hashing algorithm](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) (custom to avoid allocations)
- Tests against the logs to show there are no conflicts and tests showing basic json validation (there are edge cases here that dont get handled specifically if there is an array this code just makes the assumption that the array is going to be valid)
- Benchmark against all the test logs to show performance

After everything looks good here I am going to create custom structure hashers for other log types, (xml, syslog, clf ...) as well as a generic best guess hashing function. 

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/logtypedetectionprocessor/fingerprint_test.go
```diff
@@ -0,0 +1,48 @@
...
+		fp := fingerprintLog(body)
+		require.NotZero(t, fp, "no fingerprint for %s: %s", logType, body)
+
+		if prev, ok := seen[fp]; ok {
+			require.Equal(t, prev, logType, "fingerprint %x collides between %s and %s", fp, prev, logType)
+			continue
+		}
+		seen[fp] = logType
+	}
+}
```
colelaven, line 48: it would be nice to add a test(s) either in this file or `fingerprint_json_test.go` that validate the same fingerprint is generated from 2 logs with the same keys/structure but differing values

processor/logtypedetectionprocessor/fingerprint.go
```diff
@@ -0,0 +1,54 @@
...
+	data = strings.TrimSpace(data)
+	if len(data) < 2 {
+		return 0
+	}
+
+	first, last := data[0], data[len(data)-1]
+	if (first == '{' && last == '}') || (first == '[' && last == ']') {
+		if res := fingerprintJSON(data); res != 0 {
+			return res
+		}
```
colelaven, line 30: This seems cleaner and equivalent with the current implementation. Are you anticipating running "json-looking" data through more fingerprint functions if the fingerprintJSON returns 0?
 
```suggestion
		return fingerprintJSON(data)
```

processor/logtypedetectionprocessor/fingerprint.go
```diff
@@ -0,0 +1,54 @@
...
+
+	return 0
+}
+
+const (
+	fnvOffsetBasis = 14695981039346656037
+	fnvPrime       = 1099511628211
+)
+
+// foldFNVHashString mixes s into the running FNV-1a digest h.
```
colelaven: Did you mean to keep `s` and `h` in the comment here or did those var names change at some point?

processor/logtypedetectionprocessor/testdata/jsonLogs.csv
```diff
@@ -0,0 +1,501 @@
+type,log
```
colelaven, line 1: any reason to keep this file `.txt` instead of `.csv`?

colelaven (commented):
Some nits from a few code passes. It is mostly looking good 👍

processor/logtypedetectionprocessor/fingerprint.go
dyl10s: ya if the fingerprint is 0 it will run the other functions.

processor/logtypedetectionprocessor/fingerprint.go
dyl10s: oh yes that was some slop I reworded

processor/logtypedetectionprocessor/fingerprint.go
colelaven: Sounds good we can keep this as is then 👍

colelaven (approved):
looking good, thanks for adding those tests, they're helpful as examples for understanding this processor 👍

PR #676, opened 2026-08-04
feat: add RBAC support for ingestion labels in SecOps Exporter

TimDiam0nd:
### Proposed Change
This pull request adds support for allowed ingestion labels to be used for RBAC.

By default, the legacy gRPC/malachite API endpoint allows all ingestion labels to be used for data RBAC. However, with the new v1alpha API, a `rbac_enabled` key with a bool value must be provided alongside each label. This does exist in the proto:
https://github.com/observIQ/bindplane-otel-contrib/blob/48be36371444ddda65121e954705c5ef2aa07aff/exporter/chronicleexporter/protos/chronicle_http.proto#L44

However, this defaults to false when the struct is initialised, and no way of setting this to true currently exists.

Because the gRPC/malachite API endpoint defaults to true for RBAC enabled, this may have been overlooked and not previously considered.

Related discussion: https://security.googlecloudcommunity.com/google-security-operations-2/data-rbac-not-working-for-v1alpha-chronicle-api-604

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


exporter/chronicleexporter/marshal.go
jsirianni, line 356: Function `getRawField` needs to be extended to support looking up the new field. This function improves performance significantly as it allows us to bypass calls to the `ottl` package. 

Another `case` that checks for `chronicle_rbac_enabled` should improve performance drastically.

exporter/chronicleexporter/marshal.go
```diff
@@ -287,14 +289,23 @@ func (m *protoMarshaler) aggregateIngestionLabels(logRecord plog.LogRecord) map[
...
 			mergedLabels[cleanKey] = value.AsString()
 		}
 		return true
 	})
 	return mergedLabels
 }
 
+func (m *protoMarshaler) getRBACEnabled(ctx context.Context, logRecord plog.LogRecord, scope plog.ScopeLogs, resource plog.ResourceLogs) bool {
+	// check for attributes in attributes["chronicle_rbac_enabled"]
+	rbacAttr, err := m.getRawField(ctx, chronicleRBACField, logRecord, scope, resource)
```
jsirianni, line 303: This error should be handled. It can return useful information, such as invalid type for the rbac enabled value (bool vs string).

exporter/chronicleexporter/marshal.go
```diff
@@ -287,14 +289,23 @@ func (m *protoMarshaler) aggregateIngestionLabels(logRecord plog.LogRecord) map[
...
+	if err == nil && rbacAttr != "" {
+		return strings.ToLower(rbacAttr) == "true"
```
jsirianni: I think we should prefer `strconv.ParseBool` to ToLower.

jsirianni (changes requested):
Hi @TimDiam0nd thank you for the contribution. I left a couple comments.

TimDiam0nd:
Hi @jsirianni 
Apologies about that, has been a while since ive touched Go. Ive tried to resolve all concerns, thanks a lot.

exporter/chronicleexporter/marshal.go
```diff
@@ -356,6 +383,19 @@ func (m *protoMarshaler) getRawField(ctx context.Context, field string, logRecor
...
+			switch rbacEnabled := rbacEnabled.(type) {
+			case string:
+				return rbacEnabled, nil
+			case bool:
+				return strconv.FormatBool(rbacEnabled), nil
+			default:
+				return "", fmt.Errorf("unsupported chronicle rbac enabled type: %T", rbacEnabled)
+			}
+		}
+		return "", nil
```
jsirianni: Instead of copying all attributes, use `logRecord.Attributes().Get(chronicleRBACAttribute)`. This does go against the established pattern in this switch. Ill evaluate using `Get` for the other cases in a separate PR.

```suggestion
              if rbacEnabled, ok := logRecord.Attributes().Get(chronicleRBACAttribute); ok {
                      switch rbacEnabled.Type() {
                      case pcommon.ValueTypeStr:
                              return rbacEnabled.Str(), nil
                      case pcommon.ValueTypeBool:
                              return strconv.FormatBool(rbacEnabled.Bool()), nil
                      default:
                              return "", fmt.Errorf("unsupported chronicle rbac enabled type: %s", rbacEnabled.Type())
                      }
              }
              return "", nil
```

exporter/chronicleexporter/marshal.go
```diff
@@ -307,15 +329,20 @@ func (m *protoMarshaler) getGRPCIngestionLabels(logRecord plog.LogRecord) []*api
 	return labels
 }
 
-func (m *protoMarshaler) getHTTPIngestionLabels(logRecord plog.LogRecord) map[string]*api.Log_LogLabel {
+func (m *protoMarshaler) getHTTPIngestionLabels(ctx context.Context, logRecord plog.LogRecord, scope plog.ScopeLogs, resource plog.ResourceLogs) (map[string]*api.Log_LogLabel, error) {
 	mergedLabels := m.aggregateIngestionLabels(logRecord)
+	rbacEnabled, err := m.getRBACEnabled(ctx, logRecord, scope, resource)
+	if err != nil {
+		return nil, err
```
jsirianni: Returning this error will cause the log record(s) to be dropped.  I think it would be preferable to log the error as a warning instead of returning it. This means we can remove the `error` return value from the function signature as well.

exporter/chronicleexporter/marshal.go
jsirianni: This change will likely require test updates.

exporter/chronicleexporter/marshal.go
```diff
@@ -307,15 +329,20 @@ func (m *protoMarshaler) getGRPCIngestionLabels(logRecord plog.LogRecord) []*api
 	return labels
 }
 
-func (m *protoMarshaler) getHTTPIngestionLabels(logRecord plog.LogRecord) map[string]*api.Log_LogLabel {
+func (m *protoMarshaler) getHTTPIngestionLabels(ctx context.Context, logRecord plog.LogRecord, scope plog.ScopeLogs, resource plog.ResourceLogs) (map[string]*api.Log_LogLabel, error) {
 	mergedLabels := m.aggregateIngestionLabels(logRecord)
+	rbacEnabled, err := m.getRBACEnabled(ctx, logRecord, scope, resource)
```
jsirianni: If the length of `mergedLabels` is 0`, rbacEnabled is not useful as it is only used if there are mergedLabels (See the for loop toward the end of this function).

This function already had this issue, it should be returning early if there are no labels. Can you add a check after the call to `m.aggregateIngestionLabels`?

```suggestion
    mergedLabels := m.aggregateIngestionLabels(logRecord)
    if len(mergedLabels) == 0 {
          return labels, nil
    }

    rbacEnabled, err := m.getRBACEnabled(ctx, logRecord, scope, resource)
```

This fixes the existing issue and avoids looking up rbacEnabled.

jsirianni (changes requested):
Hi @TimDiam0nd thanks for following up and iterating with me.

I left a couple more comments. We will also need to document the rbac enabled config option in the components README.

tbm48813:
Hi Tim,
  My name is Travis, I am doing QA for this Bindplane Collector merge and I was curious if we could meet to discuss? I just had a couple workflow questions on the use case if possible.

TimDiam0nd:
Hey @jsirianni, just wanted to politely bump this

TimDiam0nd:
Hey @jsirianni apologies to bug you again about this

PR #677, opened 2026-08-05
chore: Add telemetry to log type detection

dyl10s:
Adds a counter `otelcol_log_type_detection_runs` to track how many times the "detection" code will have run. Same thing as every time there is a new fingerprint. This will help when testing real sources can be used to see how well the fingerprinting works.

Also sets up the basic fingerprint map, this will be expanded on later on with things like, max size. 

##### Checklist
- [x] Changes are tested
- [x] CI has passed


processor/logtypedetectionprocessor/metadata.yaml
```diff
@@ -7,3 +7,14 @@ status:
...
   codeowners:
     active: [observIQ/ai]
+
+telemetry:
+  metrics:
+    log_type_detection_runs:
+      enabled: true
+      description: "Number of log type detections run, one per newly seen log structure."
+      stability: alpha
+      unit: "{detections}"
```
colelaven: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units
```suggestion
      unit: "{detection}"
```

colelaven (changes requested):

processor/logtypedetectionprocessor/processor_test.go
```diff
@@ -0,0 +1,149 @@
...
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	for range 16 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-start
+			_, err := p.processLogs(context.Background(), logsFromBodies(`{"a":1,"b":"x"}`))
+			require.NoError(t, err)
+		}()
```
colelaven:
```suggestion
		wg.Go(func() {
			<-start
			_, err := p.processLogs(context.Background(), logsFromBodies(`{"a":1,"b":"x"}`))
			require.NoError(t, err)
		})
```

processor/logtypedetectionprocessor/processor.go
```diff
@@ -50,9 +61,30 @@ func (p *logTypeDetectionProcessor) processLogs(_ context.Context, ld plog.Logs)
...
+							return logType, nil
+						},
+					)
+					if err != nil {
+						return ld, err
+					}
+					logType = newLogType.(string)
+				}
 				logRecord.Attributes().PutStr("fingerprint", strconv.FormatUint(fingerprint, 16))
+				logRecord.Attributes().PutStr("logType", logType.(string))
```
colelaven: Might be worth gating the adding of these 2 attributes only if they are non-default, ie only add logType if logType is not "" and fingerprint if is not 0

dyl10s: ya I can do that, fingerprint is going to go away eventually its mainly just there for debugging

colelaven (approved):

PR #697, opened 2026-08-10
fix(lookupprocessor): bound in-memory cache and stop serializing reads

eKuG:
### Proposed Change

Fixes PIPE-1391. Two defects in `LookupCache`:

1. **The in-memory cache had no bound.** An entry was only reclaimed when the same key was looked up again after expiring, so a key that never recurred stayed resident for the life of the process. Reproduced: 400k unique keys left 400k entries resident, including 200k expired ones (~85MB per 200k entries).
2. **Every lookup took an exclusive lock.** `get` deleted expired entries inline, so it held `memMu` exclusively and every cache hit serialized against every other. Reproduced with a parallel-hit benchmark: 129 ns/op at 1 CPU degrading to 238 ns/op at 8 CPUs.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


eKuG:
* **#697** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/697?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/697?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

processor/lookupprocessor/cache_test.go
```diff
@@ -110,13 +111,161 @@ func TestLookupCache_InMemory_ExpiredEntryDeleted(t *testing.T) {
...
+
+	var wg sync.WaitGroup
+	for g := 0; g < 8; g++ {
+		wg.Add(1)
+		go func(g int) {
+			defer wg.Done()
+			for i := 0; i < 2000; i++ {
+				// Overlapping key space mixes hits, misses, and evictions.
+				_, err := c.Lookup(context.Background(), fmt.Sprintf("key-%d", (g*500+i)%256))
+				require.NoError(t, err)
```
Dylan-M: `require` in the goroutine is a testify misuse. `FailNow` has to run on the test goroutine, so from a child goroutine it just kills that one and can hang or mislead the run. `keyedSource` never errors, so it doesn't fire today. I'd still use `assert` here and check the error after `wg.Wait()`.

processor/lookupprocessor/processor.go
```diff
@@ -92,10 +98,16 @@ func (p *lookupProcessor) start(ctx context.Context, host component.Host) error
 		ttl = p.cfg.CacheTTL
 	}
 
+	maxEntries := defaultCacheMaxEntries
```
Dylan-M: Defaulting is split between here and `NewLookupCache`. `maxEntries` is guarded in both (`start`'s `if cfg > 0` and `NewLookupCache`'s `if < 1`). `ttl` is guarded only here, and `NewLookupCache` doesn't self-guard it. This `start` `maxEntries` block is also the 0%-covered branch. `NewLookupCache`'s fallback reads 100%.

Claude recommended this path: move the `ttl <= 0` fallback into `NewLookupCache`, next to its existing `maxEntries < 1` one. Then drop both clamps here and pass `cfg.CacheTTL` and `cfg.CacheMaxEntries` straight through. That gives one owner for defaulting, makes `NewLookupCache` correct regardless of caller, and lands the logic in the function that already reads 100% (a one-line test passing `ttl: 0` keeps it there).

Dylan-M (changes requested):
Details inline.

Dylan-M (approved):

PR #698, opened 2026-08-10
feat: XML fingerprinting for log types

dyl10s:
### Proposed Change
- XML Fingerprinting function determined if we start with `<` and end with `>`
- Fixed a race where log type detection could run multiple times for a hash
- Moved some variables out of the json specific fingerprint file because they can be shared across fingerprint files

The detection here became pretty basic after trying a few different things. Just hash the tag names and attributes in order and ignore tag names if they are one after another (helps with arrays) ex:
```xml
<EventData>
    <Data>Test</Data>
    <Data>Information</Data>
</EventData>
```
would be the same as
```xml
<EventData>
    <Data>Hello</Data>
</EventData>
```

Blitz sending all different types of windows events ends up with 7 unique hashes
image

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/logtypedetectionprocessor/fingerprint_json.go
```diff
@@ -14,9 +14,6 @@
 
 package logtypedetectionprocessor
 
-// maxDepth bounds object nesting so the bracket stack stays on the stack.
-const maxDepth = 64
-
 var space, structural [256]bool
```
colelaven: can you move these vars to `fingerprint.go` as well as they are used by both json & xml fingerprint funcs?

colelaven (changes requested):

processor/logtypedetectionprocessor/fingerprint_json.go
dyl10s: moved some stuff around and did some renaming, resynced the stack 🚀

processor/logtypedetectionprocessor/processor.go
colelaven: unrelated nit - no need to check for `> 0` as `== 0` is already checked above
```suggestion
logRecord.Attributes().PutStr("fingerprint", strconv.FormatUint(fingerprint, 16))
```

colelaven (approved):
👍

PR #702, opened 2026-08-11
fix(lookupprocessor): build the CSV index off-lock and publish via atomic pointer (PIPE-1179)

Dylan-M:
### Proposed Change

Part of PIPE-1179.

`CSVFile.Load` held a write lock across `os.Open`, `csv.ReadAll`, and the reindex. `CSVFile.Lookup` took a read lock and runs per record on the receiver handler goroutine, so every reload blocked all in-flight lookups until it finished. The reload fires on a hard-coded 60s ticker whether or not the file changed.

The index is now built into a local map and published with a single `atomic.Pointer` store. Readers observe either the complete previous index or the complete new one, never a partial rebuild, and a reload no longer blocks a lookup.

Measured through a collector using the following Claude driven harness method: blitz over OTLP gRPC into `otlp -> lookup -> batch -> nop`, a 1M-row 70.5 MB CSV, ~19,000 records/sec, 10 minutes per run, 11 reloads each.

| Metric               | Before      | After      |
|----------------------|-------------|------------|
| gRPC calls           | 127,437     | 130,735    |
| Mean latency         | 4.85 ms     | 3.50 ms    |
| Calls over 100 ms    | 120         | 2          |
| Calls over 500 ms    | 120         | 0          |
| Calls over 1000 ms   | 120         | 0          |
| Worst latency bucket | <= 5000 ms  | <= 250 ms  |

All of the slow calls from prior to the change happened during a reload with 10 reloads and blitz's 12 output workers blocking on the write lock together each time. Throughput is only 2.6% higher, since a stall that takes about 2% of the clock time can't move it much. This fixes latency, not throughput.

`batch` sits after `lookup` so the receive path stays synchronous through it. Batch first would hide the stall from request latency.

The benchmark runs `cache_enabled: false`. With the cache on and few distinct keys, hits never reach `CSVFile.Lookup` and the stall is invisible. The reporting customer has roughly 1M distinct hosts and misses constantly.

Manual:

1. Check out this branch in a `bindplane-otel-contrib` clone: `gh pr checkout 702`
   ```
   ```
3. `make agent`
4. Build a CSV big enough that a reload takes about a second:
   ```
   { echo "host,owner,env"; for i in $(seq 1 1000000); do echo "host-$i,owner-$((i%997)),production"; done; } > /tmp/assets.csv
   ```
5. Run the collector with a lookup over it:
   ```yaml
   receivers:
     otlp:
       protocols:
         http:
           endpoint: 127.0.0.1:4318
   processors:
     lookup:
       context: attributes
       field: host
       csv: /tmp/assets.csv
   exporters:
     debug:
       verbosity: detailed
   service:
     telemetry:
       logs:
         level: debug
     pipelines:
       logs:
         receivers: [otlp]
         processors: [lookup]
         exporters: [debug]
   ```
6. Once `source loaded/refreshed` appears, send a record every second and watch how long each request takes:
   ```
   while true; do
     /usr/bin/time -f '%e s' curl -s -o /dev/null -X POST http://127.0.0.1:4318/v1/logs \
       -H 'Content-Type: application/json' \
     sleep 1
   done
   ```
7. Every request should stay in the low milliseconds, including at the 60-second reload boundaries. Confirm the debug output shows `owner` and `env` added to each record throughout.
8. To see the behavior this replaces, restore the lock: add a `sync.RWMutex` to `CSVFile`, take `Lock()` at the top of `Load` and `RLock()` in `Lookup`, rebuild, and repeat. One request per minute now takes roughly a second while the reload holds the lock.

Automated: `go test -race ./...` in `processor/lookupprocessor`. The race detector matters, since this swaps a mutex for an `atomic.Pointer` on a path read by every record.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/processor/lookupprocessor/csv_reload_stall_test.go
+++ b/processor/lookupprocessor/csv_reload_stall_test.go
@@ -1,0 +1,108 @@
+// Copyright  observIQ, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lookupprocessor
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// writeStallCSV writes a small host-keyed CSV and returns its path.
+func writeStallCSV(t *testing.T, rows int) string {
+	t.Helper()
+
+	path := filepath.Join(t.TempDir(), "assets.csv")
+	f, err := os.Create(path)
+	require.NoError(t, err)
+	defer func() { require.NoError(t, f.Close()) }()
+
+	_, err = fmt.Fprintln(f, "host,owner,env")
+	require.NoError(t, err)
+	for i := 0; i < rows; i++ {
+		_, err = fmt.Fprintf(f, "host-%04d,owner-%d,production\n", i, i)
+		require.NoError(t, err)
+	}
+	require.NoError(t, f.Sync())
+
+	return path
+}
+
+// TestReloadDoesNotBlockLookups holds a reload in flight and asserts that a
+// lookup completes against the still-published old index rather than waiting for
+// the reload to finish.
+//
+// The index is published through an atomic pointer, so a reload builds the new
+// map off to the side and swaps it in with one Store. readHook pauses the reload
+// after it has built the new index but before that Store, which is the exact
+// window a concurrent lookup would have blocked in under a lock held across the
+// whole reload. A lookup issued during that window must still return, since it
+// reads the old index and never takes the reload's path.
+func TestReloadDoesNotBlockLookups(t *testing.T) {
+	path := writeStallCSV(t, 100)
+	csvFile := NewCSVFile(path, "host")
+	require.NoError(t, csvFile.Load())
+
+	// Confirm the key resolves off the initial index before the reload starts.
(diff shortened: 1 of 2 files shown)


Dylan-M:
* **#704** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#703** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#702** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

processor/lookupprocessor/csv_reload_stall_test.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// TestReloadDoesNotStallLookups measures per-Lookup latency on a steady reader
+// goroutine while a reload runs concurrently, mirroring the data path
+// (ConsumeLogs -> addLookupValues -> Lookup) racing the loadSource ticker.
+//
+// The write lock in Load() is held across os.Open + csv.ReadAll + indexRecords,
+// so every concurrent Lookup blocks for the full reload duration.
+func TestReloadDoesNotStallLookups(t *testing.T) {
+	const rows = 1_000_000
```
Caleb-Hurshman: A little worried about the time/memory impact of adding this test with 1 million rows (and 1 million write syscalls) to the suite.
What sort of performance has this had locally for you, and is there a more efficient way we could test these changes?

Caleb-Hurshman (approved):

Dylan-M:
### Merge activity

* **Aug 18, 12:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702).
* **Aug 18, 12:23 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702).

PR #703, opened 2026-08-11
fix(lookupprocessor): skip CSV reloads when the file is unchanged and make the interval configurable (PIPE-1179)

Dylan-M:
### Proposed Change

Part of PIPE-1179.

The reload re-read and re-indexed the whole CSV every 60s even when the file had not changed. `Load` now stats first and skips the read when modification time and size both match the last successful load. A static file is therefore read once rather than once a minute.

A file rewritten in place can be read mid-write, but still parseable. `Load` re-stats after reading and reads again if the file moved.

`reload_interval` is now configurable, defaulting to the historical 60s.

Stamp comparison uses `time.Equal` rather than `==`. On a `time.Time`, `==` also compares the monotonic reading and the location pointer, so it can report a difference for the same instant and silently defeat the skip.

Measured with the same harness, 10 minutes per run, ~11.5M records each, against PR 1 as the baseline.

Static CSV:

| Metric            | Baseline | With skip |
|-------------------|----------|-----------|
| Load calls        | 11       | 11        |
| Actual file reads | 11       | 1         |
| Collector CPU     | 514.91 s | 372.59 s  |

CSV rewritten atomically every 240s, so 10 load calls meet 2 genuine changes:

| Metric            | Baseline | With skip |
|-------------------|----------|-----------|
| Load calls        | 11       | 11        |
| Actual file reads | 11       | 3         |
| Collector CPU     | 519.32 s | 369.42 s  |

27.6% less CPU on the static file and 28.9% on the rewritten one. The rewritten-file run is the one that matters for correctness: 3 reads for 1 initial load plus 2 real changes shows the skip still reloads when the file turns over. A static-only benchmark cannot distinguish a working skip from one that never reloads at all, since the broken version would post better numbers.

`loadSource` logged `source loaded/refreshed` on every successful `Load`, which would now claim a reload happened on a skip. It logs `source reload check complete` instead, and `CSVFile` logs the skip and the retry itself.

Manual:

1. Check out this branch in a `bindplane-otel-contrib` clone: `gh pr checkout 703`
   ```
   ```
3. `make agent`
4. A small CSV is enough here, since this is about when the file is read rather than how long a read takes:
   ```
   printf 'host,owner,env\nhost-42,alice,production\n' > /tmp/assets.csv
   ```
5. Run the collector with a short interval so cycles are quick to watch:
   ```yaml
   receivers:
     otlp:
       protocols:
         http:
           endpoint: 127.0.0.1:4318
   processors:
     lookup:
       context: attributes
       field: host
       csv: /tmp/assets.csv
       reload_interval: 10s
   exporters:
     debug:
       verbosity: detailed
   service:
     telemetry:
       logs:
         level: debug
     pipelines:
       logs:
         receivers: [otlp]
         processors: [lookup]
         exporters: [debug]
   ```
6. Watch the log for 30 seconds. The first cycle reads the file. Every cycle after it logs `csv unchanged, skipping reload` with the path, confirming the interval still fires while the read is avoided.
7. Change the file's contents: `printf 'host,owner,env\nhost-42,bob,staging\n' > /tmp/assets.csv`. Within one interval the skip line stops and the file is read again.
8. Confirm the new values reach records. Send one before and one after the change and compare the debug output:
   ```
   curl -s -X POST http://127.0.0.1:4318/v1/logs -H 'Content-Type: application/json' \
   ```
   `owner` should go from `alice` to `bob` and `env` from `production` to `staging`. This is the check that matters most: a skip that never reloads would keep serving `alice` forever and would otherwise look like a successful optimization.
9. Confirm the interval is validated. Set `reload_interval: -1s` and restart; the collector refuses to start with `reload_interval must not be negative`.

Automated: `go test -race ./...` in `processor/lookupprocessor`. `TestLoadSkipsUnchangedFile` asserts the file is not re-read and the published index pointer is unchanged, `TestLoadPicksUpChangedFile` asserts a changed file is re-read, and `TestLoadRetriesWhenFileChangesMidRead` drives a rewrite between the read and the re-stat to confirm the retry fires exactly once.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/processor/lookupprocessor/csv.go
+++ b/processor/lookupprocessor/csv.go
@@ -21,6 +21,9 @@ import (
 	"fmt"
 	"os"
 	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


Dylan-M:
* **#704** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#703** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#702** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

processor/lookupprocessor/csv.go
```diff
@@ -47,30 +50,106 @@ type index map[string]map[string]string
...
+	// reads counts completed file reads, and readHook runs between a read and
+	// the re-stat that follows it. Both exist for tests; readHook is nil in
+	// production.
+	reads    atomic.Int64
+	readHook func()
+}
+
+// fileStamp identifies a file version cheaply. It relies on a write advancing the
+// modification time, which holds on any filesystem storing mtime at sub-second
+// resolution. A rewrite that deliberately restores the previous timestamp and
```
Caleb-Hurshman: Is it possible to get two writes within the same tick, and thus the same modTime? If so, 2 writes with equal size and modTime would indefinitely skip on every Load(), right?

Maybe that's a far fetched state, but if it's possible it seems like we should guard against it.

Caleb-Hurshman (approved):

Dylan-M:
### Merge activity

* **Aug 18, 12:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703).
* **Aug 18, 12:24 PM UTC**: [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703) rebased this pull request as part of a merge.
* **Aug 18, 12:44 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703).

PR #704, opened 2026-08-11
fix(lookupprocessor): bypass the cache for CSV sources and warn on cache settings (PIPE-1179)

Dylan-M:
### Proposed Change

Fixes PIPE-1179.

A CSV index is already fully resident in memory, so fronting it with `LookupCache` cannot reduce work. `start` now uses a CSV source directly and wraps only Redis and API.

Measured per lookup, 8 parallel readers, against a 100k-row CSV:

| Workload     | Direct      | Through cache | Increase |
|--------------|-------------|---------------|----------|
| 10 keys      | 57.17 ns/op | 350.0 ns/op   | +512%    |
| 100,000 keys | 75.82 ns/op | 350.6 ns/op   | +362%    |

The overhead is not lock contention. #697 fixed that: `get` now takes only a read lock and never mutates. What remains is structural, a `fmt.Sprintf` to build the cache key, a clock read, and a second map probe, spent to avoid a lookup that already costs 57ns.

Cached latency is flat at ~350ns across both key counts, so no key distribution makes the cache pay for itself here.

This is a bypass rather than a validation error on purpose. `createDefaultConfig` enables the cache, so rejecting `cache_enabled` alongside `csv` would fail every existing CSV configuration at startup.

Setting `cache_enabled`, `cache_ttl`, or `cache_max_entries` with a CSV source now warns that the setting is ignored and becomes a configuration error in a future release. It fires on any explicit value including `false`, so a configuration that will error later is told now rather than surprised on upgrade.

Detecting that needs `Unmarshal` to record which keys the configuration actually contained. The factory fills all three defaults, so a value check cannot tell a user's choice from a default and would warn for every existing CSV user.

Manual:

1. Check out this branch in a `bindplane-otel-contrib` clone: `gh pr checkout 704`
   ```
   ```
3. `make agent`
4. `printf 'host,owner,env\nhost-42,alice,production\n' > /tmp/assets.csv`
5. Start with a config that sets a cache key alongside `csv`:
   ```yaml
   receivers:
     otlp:
       protocols:
         http:
           endpoint: 127.0.0.1:4318
   processors:
     lookup:
       context: attributes
       field: host
       csv: /tmp/assets.csv
       cache_enabled: true
   exporters:
     debug:
       verbosity: detailed
   service:
     telemetry:
       logs:
         level: debug
     pipelines:
       logs:
         receivers: [otlp]
         processors: [lookup]
         exporters: [debug]
   ```
6. At startup the collector warns once: `cache settings do not apply to a csv source and are ignored; setting them with 'csv' becomes a configuration error in a future release`, with a `keys` field naming `cache_enabled`. Confirm lookups still work by sending a record and seeing `owner` and `env` added:
   ```
   curl -s -X POST http://127.0.0.1:4318/v1/logs -H 'Content-Type: application/json' \
   ```
7. Add `cache_ttl: 1m` and `cache_max_entries: 10` and restart. The `keys` field now names all three, still in one warning rather than three.
8. Set `cache_enabled: false` and restart. It still warns, because the key does not apply to `csv` whatever its value, and a config that will error in a future release should hear about it now rather than on upgrade.
9. Remove all three keys and restart. No warning appears, which is the case that matters most given the factory fills all three defaults.

Automated: `go test -race ./...` in `processor/lookupprocessor`. `TestCSVSourceIsNotWrappedInCache` asserts the source is a bare `CSVFile`, `TestRedisAndAPISourcesStayWrappedInCache` asserts non-CSV sources keep the cache, and `TestWarnsOnCacheKeysSetWithCSV` plus `TestNoWarningWhenCacheKeysAbsentWithCSV` cover the four warning cases above.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Dylan-M:
* **#704** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#703** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#702** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/702?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

processor/lookupprocessor/config.go
```diff
@@ -65,6 +66,43 @@ type Config struct {
...
+
+	// cacheKeysSet records which cache keys the configuration actually contained.
+	// createDefaultConfig fills all three, so a plain value check cannot tell a
+	// user's choice from a default, and warning on the default would fire for
+	// every existing CSV user.
+	cacheKeysSet map[string]bool
+}
+
+// cacheKeys are the settings that only apply to a source the cache fronts.
+var cacheKeys = []string{"cache_enabled", "cache_ttl", "cache_max_entries"}
```
Caleb-Hurshman:
```suggestion
var cacheKeys = []string{"cache_enabled", "cache_ttl", "cache_max_entries", "storage"}
```
Storage is silently ignored right now for CSV

processor/lookupprocessor/README.md
Caleb-Hurshman: It would be good to add a note to these to mention "Not used with csv" or something along those lines.

Caleb-Hurshman, line 11: This section is also a little stale after the CSV changes.

Caleb-Hurshman (approved):

Dylan-M:
### Merge activity

* **Aug 18, 12:23 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704).
* **Aug 18, 12:46 PM UTC**: [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704) rebased this pull request as part of a merge.
* **Aug 18, 1:07 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/704).

PR #705, opened 2026-08-12
chore: CLF fingerprint function

dyl10s:
### Proposed Change

Adds the CLF fingerprinting function.

Because this is a standardized format things have to get a little tricky here. Common log format is fingerprinted based on the actual log information which is almost exclusivly an HTTP request. The hash is therefor based on the method and url.

In urls typically a number in the path is used to identify a resource. This is not always the case, but it is a common practice and is a safe assumption here for now so paths like "/users/1" and "/users/2" get the same fingerprint as well as "/users/ABC1" and "/users/ABC2".

My guess is most of the time we will not be able to identify the log type for CLF logs but if we eventually let users do some self assignment then this will still be needed.

Fixes BP-76

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_clf_test.go
```diff
@@ -0,0 +1,136 @@
...
+			shouldEqual: true,
+		},
+		{
+			title:       "different hashed asset name",
+			logA:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /static/css/main.7f3a1c.css HTTP/1.1" 200 10`,
+			logB:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /static/css/main.9b2e4d.css HTTP/1.1" 200 10`,
+			shouldEqual: true,
+		},
+		{
+			title:       "different query value",
```
colelaven, line 92: Is it intentional that non-numeric differing query values with the same key generate different fingerprints? It seems like it'd make sense to ignore query values no matter if they are numeric or not, otherwise 2 logs like those below would not match fingerprints.

A test like this would be nice to add assuming we want them to equal, this fails right now:
```
       {
			title:       "different query value non-numeric",
			logA:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /accounts?user=joe HTTP/1.1" 200 10`,
			logB:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /accounts?user=ja HTTP/1.1" 200 10`,
			shouldEqual: true,
		},
```

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_clf_test.go
```diff
@@ -0,0 +1,136 @@
...
+			shouldEqual: true,
+		},
+		{
+			title:       "different referer and user agent",
+			logA:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET / HTTP/1.1" 200 10 "-" "curl/8.7.1"`,
+			logB:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET / HTTP/1.1" 200 10 "http://example.com/" "Mozilla/5.0"`,
+			shouldEqual: true,
+		},
+		{
+			title:       "different http version",
```
colelaven, line 74: similar to this test, itd be nice to add a test with differeing API versions. Seems like a relatively realistic scenario:
```
		{
			title:       "different api version",
			logA:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /v1/orders HTTP/1.1" 200 10`,
			logB:        `192.168.14.77 - - [04/Aug/2026:09:12:41 +0000] "GET /v2/orders HTTP/1.1" 200 10`,
			shouldEqual: true,
		},
```

colelaven (changes requested):

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_clf_test.go
dyl10s: ya I agree and I think it can even be more generic, honestly the query strings entirely should be dropped. A request to `/accounts` vs `/accounts?user=ja` are very likely to be pointing to the same service

colelaven: Agreed, that seems like a good change to make 👍

colelaven (approved):
lgtm, may be worth rerunning the benchmarks and updating numbers, but not a big deal

PR #706, opened 2026-08-13
chore: Syslog fingerprint function

dyl10s:
### Proposed Change

This PR adds the syslog fingerprinting function. This function does not do a whole lot on its own because syslog is
mostly a wrapper around existing logs. This function does a fingerprint on the service name from the syslog
and then passes the body to the other fingerprinting functions.

Currenly the generic fingerprinting function is not implemented so the test for unique fingerprints is
commented out.

There is also a boolean to block this recursing

##### Checklist
- [x] Changes are tested
- [x] CI has passed


processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestFingerprintSyslogStructure(t *testing.T) {
+	cases := []struct {
+		title      string
+		logA, logB string
+		equal      bool
+	}{
+		{
+			title: "3164 same tag different header and message",
```
dyl10s, line 75: Fallback right now any service that same tag and a message that the fingerprint functions wont be able to fingerprint will be the same. I would expect something like this to be false eventually.

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint.go
```diff
@@ -47,9 +47,16 @@ func HashLog(data string) uint64 {
 		}
 	}
 
-	if first == '<' && last == '>' {
-		if res := fingerprintXML(data); res != 0 {
-			return res
+	if first == '<' {
+		if !skipRecursiveTypes && (data[1] >= '0' && data[1] <= '9') {
```
colelaven, line 57:
```suggestion
		if !skipRecursiveTypes && isDigit(data[1]) {
```

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
```diff
@@ -0,0 +1,210 @@
...
+	i := skipSpaces(data, 16)
+	end := endOfToken(data, i)
+	if end == i {
+		return ""
+	}
+
+	return data[skipSpaces(data, end):]
+}
+
+func isSyslog3164Timestamp(data string) bool {
```
colelaven, line 116: nit, might be worth adding a len check at the beginning of this function so no array accesses later on (like `data[15]`) can cause a panic. I see there is a check where the func is called but makes sense to move it here imo.

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
```diff
@@ -0,0 +1,210 @@
...
+		service = syslog3164Tag(msg)
+		if sep := strings.Index(msg, ": "); sep >= 0 {
+			msg = msg[sep+2:]
+		}
+	}
+	if msg == "" {
+		return 0
+	}
+
+	hash := foldFNVHashString(fnvOffsetBasis, service)
```
colelaven, line 59: edge case: if `service == ""` here, do we still want to fold it into the hash?

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
```diff
@@ -0,0 +1,210 @@
...
+	return foldFNVHashUint(hash, HashLog(msg, true))
```
colelaven: Similarly, if `HashLog(msg, true) == 0` here, do we want to fold that into the hash?

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
```diff
@@ -0,0 +1,210 @@
...
+	if i < 0 {
+		return 0
+	}
+
+	var service, msg string
+	if strings.HasPrefix(data[i:], "1 ") {
+		service, msg = syslog5424Message(data[i+2:])
+	} else {
+		msg = syslog3164Message(data[i:])
+		service = syslog3164Tag(msg)
```
colelaven: nit, any reason `syslog3164Message` can't be implemented with the same interface as `syslog5424Message` so that both service, and msg are simply returned at once?

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint.go
```diff
@@ -34,7 +34,7 @@ func init() {
 }
 
 // HashLog creates a hash based off the structure of the log.
-func HashLog(data string) uint64 {
+func HashLog(data string, skipRecursiveTypes bool) uint64 {
```
colelaven: nit, do you anticipate calling this func with `skipRecursiveTypes = true` anywhere outside this internal package? Otherwise it could be cleaner to remove that param from the exported function and call a hidden `hashLog` function which does have the bool exposed. Just would simplify the public interface a bit.

colelaven (changes requested):
Mostly nits, feel free to ignore any if you feel they aren't needed.

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
dyl10s: it gets skipped by the fold function so I don't think a check is needed here

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_syslog.go
dyl10s: I do like this added in because it is the "structure", having a 0 come back. It dosent really change a whole lot either way do I can add an if check if you feel strongly.

colelaven (approved):
ty for cleanup

PR #708, opened 2026-08-13
chore: Generic fingerprint function

dyl10s:
### Proposed Change

This is the last fingerprint function for the time being and is meant to handle free text. Two simple formats are attempted to be parsed:

- Key Value Pairs
- CSV

Key value pairs are fingerprinted based off the keys
CSV is fingerprinted off the type of the data in each column

Somewhat similar rules for CLF are used for any fallthough text. We keep around any string values but
replace any numbers/timestamps with placeholders. This is all shown in the test cases.

This also means that CLF will fall through to the generic parser and tests are updated there to reflect that.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


processor/logtypedetectionprocessor/internal/fingerprint/fingerprint.go
```diff
@@ -42,7 +42,7 @@ func HashLog(data string) uint64 {
 // not exposed via the public API.
 func fingerprint(data string, skipRecursiveTypes bool) uint64 {
 	data = strings.TrimSpace(data)
-	if len(data) < 2 {
+	if len(data) < 10 {
```
colelaven: just curious why did you chose 10 here?

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint_generic.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		fields++
+		if i >= len(data) {
+			break
+		}
+		hash = foldFNVHashByte(hash, ',')
+		i++
+	}
+
+	if fields < 4 {
```
colelaven: Why not fingerprint csvs with 2 or 3 fields?

dyl10s: I could lower this to 3 this is just an arbitrary number I picked where I would be pretty confident we would have a CSV. I just want to prevent fingerprinting an item as a csv if it is not a csv. The csv fingerprinting is very generic.

colelaven: Sounds good, I think 4 makes sense for this threshold, was mostly just wondering how this got decided 👍

processor/logtypedetectionprocessor/internal/fingerprint/fingerprint.go
dyl10s: seemed like a good number to me 🤷 

open to a different length if you have an opinion. This is something that I think we could tweak over time. The main thing I was trying to help do some prevention of was having an unparsed xml log come in with each line of the XML being its own log and causing the fingerprinting function to run a ton.

colelaven (approved):

PR #738, opened 2026-08-14
fix(gcspubsubeventreceiver,awss3eventreceiver): terminate parse loop on read errors [PIPE-1409]

Dylan-M:
### Proposed Change

Currently `line_parser.go` yields a read error and then falls through without a `continue`. The bare `for {}` loop has no safe exit for the circumstance. Once a context is cancelled, every subsequent `ReadLine` returns `context.Canceled` and the loop spins on it infinitely until systemd sends a SIGKILL.

A customer support bundle shows this as 1.1 million identical error lines from six Pub/Sub messages, around 10k per second. The reload from a rollout never finishes, so systemd kills the collector. The unacked messages then redeliver and the same objects reprocess from the start.

A read error is now terminal: the parser surfaces it once and stops.

A terminal read used to ack the message and delete the saved offset, because the worker logs a yielded error and keeps going. Every record after the break was lost. The parser now marks a terminal read, and the worker fails the object. The message stays queued and resumes from its offset.

`io.ErrUnexpectedEOF` means bytes are missing rather than unreachable, so a retry reads the same object. A truncated object goes to the dead-letter queue instead of the retry path. Inside an archive, the parser drops only that entry.

The visibility tests now run in `testing/synctest` bubbles on a fake clock, replacing the real-time sleeps.

Both receivers carry the identical loop, so both are fixed here. A refactor will happen later to combine shared code.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/receiver/awss3eventreceiver/internal/worker/json_parser.go
+++ b/receiver/awss3eventreceiver/internal/worker/json_parser.go
@@ -204,8 +204,10 @@ func (p *jsonParser) yieldArray(startOffset int64) iter.Seq2[any, error] {
 				if errors.Is(err, io.EOF) {
 					return
 				}
-				// unexpected end of file, not much we can do here
+				// The object stops part way through a record. The missing bytes
+				// were never written, so a retry reads the same thing.
 				if errors.Is(err, io.ErrUnexpectedEOF) {
+					yield(nil, ErrTruncatedObject{Err: err})
 					return
 				}
 				// unexpected error, return it
@@ -221,6 +223,12 @@ func (p *jsonParser) yieldArray(startOffset int64) iter.Seq2[any, error] {
 					return
 				}
 			}
+		}
+		// The loop above also ends when the object stops part way through, because
+		// More reports no further element either way. A complete array closes with a
+		// delimiter, so anything else here means the bytes ran out early.
+		if _, err := p.decoder.Token(); err != nil {
+			yield(nil, ErrTruncatedObject{Err: err})
 		}
 	}
 }
--- a/receiver/awss3eventreceiver/internal/worker/line_parser.go
+++ b/receiver/awss3eventreceiver/internal/worker/line_parser.go
@@ -16,6 +16,7 @@ package worker
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"iter"
@@ -51,12 +52,22 @@ func (p *lineParser) Parse(_ context.Context, startOffset int64) (logs iter.Seq2
 		for {
 			lineBytes, _, err := p.reader.ReadLine()
 			if err != nil {
-				if err == io.EOF {
+				// A decompressing reader can wrap the sentinel.
+				if errors.Is(err, io.EOF) {
 					return
 				}
-				if !yield(nil, err) {
+				// An object that stops mid-record is truncated rather than
+				// unreachable. A decompressor reports a live connection failure
+				// as that failure, so this only fires on missing bytes.
+				if errors.Is(err, io.ErrUnexpectedEOF) {
+					yield(nil, ErrTruncatedObject{Err: err})
 					return
 				}
+				// A read error is terminal. Every later ReadLine returns the same
+				// error, so continuing spins forever. It is marked so the caller
+				// fails the object instead of acking a partial read.
+				yield(nil, ErrStreamRead{Err: err})
+				return
 			}
 
 			// only yield non-empty lines
--- a/receiver/awss3eventreceiver/internal/worker/json_truncated_test.go
+++ b/receiver/awss3eventreceiver/internal/worker/json_truncated_test.go
@@ -1,0 +1,77 @@
+// Copyright observIQ, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package worker
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// TestJSONParser_TruncatedObjectRoutesToDLQ asserts a JSON array cut short reports the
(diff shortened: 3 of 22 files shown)


Dylan-M:
* **#743** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#742** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#738** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

receiver/awss3eventreceiver/internal/worker/line_parser.go
```diff
@@ -51,12 +52,14 @@ func (p *lineParser) Parse(_ context.Context, startOffset int64) (logs iter.Seq2
 		for {
 			lineBytes, _, err := p.reader.ReadLine()
 			if err != nil {
```
Caleb-Hurshman, line 54: Reviewing this on the side, so forgive me if this isn't entirely accurate, but it seemed important to flag (AI output incoming):


`receiver/awss3eventreceiver/internal/worker/line_parser.go:61` — high: terminal read error is swallowed by the caller, so a truncated object is acked as fully processed (silent, permanent data loss).

Making the read error terminal is the right call for the spin, but the consumer was written for the old "skip one bad record and keep going" contract. In worker.go:407-411, processObject logs the yielded error and continues; the sequence then ends, the partial batch is emitted, and consumeLogsFromS3Object returns nil. ProcessMessage therefore treats the record as a success: w.deleteMessage(ctx, msg, queueURL, keys, logger) (worker.go:318) deletes the SQS message and deletes the saved offset key (worker.go:485-490).

Concrete scenario: a 500 MB S3 object whose body read fails mid-stream with read: connection reset by peer (common for large GetObject bodies). Context is still live, so ConsumeLogs succeeds for the first N lines, the message is deleted, the offset is deleted, and the rest of the object is never ingested and never retried. Before this change the loop spun (bad) but the message was eventually redelivered and resumed from the saved offset, so no data was lost. Worse, if the error lands right after a batch flush, ld.LogRecordCount() == 0 → early return nil (worker.go:450) and the object is acked having emitted nothing at all, with only a single parse log error line to show for it.

Fix direction: distinguish a terminal read failure from a per-record parse failure (e.g. wrap it in a dedicated error type and have the worker return err instead of continue, so handleProcessingError preserves the message in SQS for retry and the saved offset allows resume).

Dylan-M: I wasn't accounting for the worker's skip-and-continue contract when I made the read error terminal. Terminal read failures are now marked, and the worker fails the object instead of acking a partial read. The message stays queued and resumes from the saved offset. A malformed record is still skipped.

Chasing this, I found the Avro parser loses records the same way, with no log line at all. It never checks the error that ended the scan. That one predates this PR and gets its own.

Caleb-Hurshman: I think this is fixed, but slightly over-corrected. `io.ErrUnexpectedEOF` from a truncated gzip is treated as retryable now, which will result in a new infinite loop. `json_parser.go:207` in this same package returns for this specific error.

Also, the new `IsStreamRead(err)` check makes a read error for a single object abort the whole object, which breaks `archive.go`'s per-entry skip contract.

So, we should 1. return on `io.ErrUnexpectedEOF`, and 2. route decode/corruption errors to the DLQ path (GCS already uses this for `ErrCorruptArchive` & `ErrArchiveLimitExceeded`) instead of the retry path.

receiver/awss3eventreceiver/internal/worker/json_parser.go
```diff
@@ -222,6 +224,12 @@ func (p *jsonParser) yieldArray(startOffset int64) iter.Seq2[any, error] {
 				}
 			}
 		}
+		// The loop above also ends when the object stops part way through, because
+		// More reports no further element either way. A complete array closes with a
+		// delimiter, so anything else here means the bytes ran out early.
+		if _, err := p.decoder.Token(); err != nil {
```
Caleb-Hurshman: `Token()` turns midstream errors into ErrTruncatedObject, which is a permanent DLQ verdict. The check needs to distinguish "bytes ran out" (`io.ErrUnexpectedEOF`/`io.EOF`) from every other error, which should become ErrStreamRead like the line parser does.

receiver/awss3eventreceiver/internal/worker/json_parser.go
Caleb-Hurshman: JSON path has the same infinite spin issue that we're trying to fix here, where an invalid character or a similar error will get logged and continue forever.
A syntaxError needs to be treated as unrecoverable.

receiver/awss3eventreceiver/internal/worker/line_parser.go
```diff
@@ -51,12 +52,22 @@ func (p *lineParser) Parse(_ context.Context, startOffset int64) (logs iter.Seq2
...
-				if err == io.EOF {
+				// A decompressing reader can wrap the sentinel.
+				if errors.Is(err, io.EOF) {
 					return
 				}
-				if !yield(nil, err) {
+				// An object that stops mid-record is truncated rather than
+				// unreachable. A decompressor reports a live connection failure
+				// as that failure, so this only fires on missing bytes.
+				if errors.Is(err, io.ErrUnexpectedEOF) {
```
Caleb-Hurshman: New finding:  An interrupted download is misclassified as a truncated object, and the S3 worker then silently acks and loses the rest of the object.
A complete S3 object whose download broke half-way is permanently acked with only its first half ingested, with nothing but a parse log error line.

Caleb-Hurshman: Also exists on the GCS side. In Claude's words:

IsTruncatedObject now feeds dlqErrorKindUnsupportedFile, a transient short read on a fully-uploaded object nacks with deadline 0 (immediate redelivery, no backoff) and burns through redeliveries into the DLQ, losing everything after the break. Truncation can only be asserted safely by comparing bytes read against the object's known size, not from ErrUnexpectedEOF alone.

Caleb-Hurshman (approved):
Approved, all findings addressed here or later in the stack

Dylan-M:
### Merge activity

* **Aug 25, 7:39 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/738).

PR #741, opened 2026-08-14
refactor(gcspubsubeventreceiver,awss3eventreceiver): share the streaming parser stack via internal/blobstream [PIPE-1411]

Dylan-M:
### Proposed Change

The two receivers carried near-identical copies of the worker and parser code. That is why the cancellation defect appeared in both.

`internal/blobstream` now holds the streaming parser stack, and both receivers use it. `internal/blobconsume` could not host it. That package takes the whole object as a `[]byte`, so it has no streaming and no mid-object resume. The byte-offset checkpoint depends on both.

S3 gains behavior from the shared code. It now detects gzip and format from content rather than from the object name or the `Content-Encoding` label. An object named `.txt` that holds a JSON array therefore parses as JSON rather than as text. It also gains archive support, streaming codecs, and archive-aware offsets.

This change reconciles three divergences, each a check that only one copy held. S3 now rejects `max_logs_emitted` of zero, which previously flushed a batch on every record. `AppendLogBody` checks the type assertion. The JSON parser stops on a decode error. A `json.Decoder` cannot resync, so the loop otherwise spins.

`s3event.parse_errors` is new. The shared code skips unparseable records, and S3 had no counter for it.

Only `worker.go` and the offset key still differ between the two packages. This is intentional, as those hold Pub/Sub and SQS lifecycle code.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

PR #740, opened 2026-08-14
fix(gcspubsubeventreceiver,awss3eventreceiver): run checkpoint, ack, and nack on a detached context [PIPE-1410]

Dylan-M:
### Proposed Change

Every wind-down operation took the context that had just been cancelled. That covers both checkpoint saves, the offset delete, the ack, and both nack paths. None of them ran. A customer support bundle shows `failed to ack message` 25 seconds after a config push.

Recovery then happened by accident. The ack deadline lapsed, the message redelivered, and work resumed from the last saved batch offset.

Those operations now run on a context detached from the cancellation, with a 30 second timeout. The worker nacks the message at once instead of waiting for its deadline.

The worker also flushes and checkpoints the partial batch before the cancellation propagates. This is only safe because the parser now drops a record that a broken stream cut short. A batch therefore ends on a record boundary, and the checkpoint stays accurate.

Cancellation is never a DLQ condition. The check is explicit, because the S3 classifiers match on error text. Cancellation also no longer counts as a failure.

The offset load still uses the caller context. It starts work rather than ending it, so a cancelled read aborts instead of loading unused state.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

PR #739, opened 2026-08-14
fix(gcspubsubeventreceiver,awss3eventreceiver): drop records truncated by a broken stream [PIPE-1415]

Dylan-M:
### Proposed Change

`bufio.Reader.ReadLine` clears the read error when it already holds bytes. A stream that breaks mid-record then resembles an unterminated final record, so the parser emits it whole.

The resume position also advances past those bytes. Redelivery starts mid-record and emits the tail as a second record. In testing, one line arrived as `"lin"` and then as `"e-000145-padding..."` across two collector lifetimes. No error appears in the logs.

`io.EOF` separates the two cases. At end of stream the trailing bytes are a complete record, so the parser emits them. Under any other error they are a fragment, so the parser drops them. `ReadLine` discards that signal, so the parser now reads through `ReadSlice`.

The parser also tracks the position of the last emitted record. The reader position moves past dropped bytes, and the checkpoint must not follow it.

Both receivers carry the same loop.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

PR #742, opened 2026-08-14
feat(gcspubsubeventreceiver,awss3eventreceiver): add raw and include_log_record_original options [PIPE-1412]

Dylan-M:
### Proposed Change

Neither receiver had an option for raw or parsed bodies. Avro OCF and JSON both parse automatically, so the body arrives as a structure.

That breaks SecOps ingestion. The exporter runs with `raw_log_field: body`, and its parser expects the original vendor text.

`raw` emits the original text as the body instead of a parsed structure. It skips parser selection and still runs the existing content check, so an image or a PDF is rejected rather than emitted as garbled text.

Avro OCF holds no original text. Under `raw` the parser emits the JSON encoding of each record as a string body.

`include_log_record_original` writes the original text to the `log.record.original` attribute and leaves the body alone. It applies to a raw body and a parsed body alike.

Both options default to false, so existing configurations parse as before.

For JSON, the original text is the element's exact bytes. For Avro it is the same JSON encoding used for a raw body.

The options also apply to archive entries. Each entry runs format detection with a new stream, so the options must carry across that boundary.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


Dylan-M:
* **#743** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#742** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/742?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/742?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#738** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/738?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

Caleb-Hurshman (commented):
No live issues, only two latent findings to document/clean-up:

Archive temp file leaks if the iterator is never ranged. Records() materializes a zip/7z to a temp file before returning, but the only cleanup is a defer inside the returned closure. Probed: called Records() on testdata/logs.7z with a scratch temp dir, abandoned the iterator, and a blobstream-7z-* file survived. The file can be as large as maxTotalBytes. Neither receiver hits this today — both range immediately — but RecordProducer is now a shared API with an undocumented "you must drive the iterator" contract, and calling Records() twice leaks the first backend too. Opening lazily inside the closure, or adding a Close(), removes it.

RawTruncated fails open when the sizes aren't comparable. It compares the raw byte count against Content-Length, but the raw counter sits below decompression while Size may describe a differently-encoded payload. If the transport transparently decompresses (S3 serving a Content-Encoding: gzip object), the counter tallies decompressed bytes against a compressed length and a genuinely short download reads as complete. Same silent false when Size <= 0 — documented for GCS transcoding, where Attrs.Size is -1. Fail-open is the right default, but there's no log distinguishing "verified complete" from "couldn't judge."

Caleb-Hurshman (approved):

Dylan-M:
### Merge activity

* **Aug 26, 7:06 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/742).

PR #743, opened 2026-08-14
feat(gcspubsubeventreceiver,awss3eventreceiver): read NDJSON as a JSON value sequence [PIPE-1413]

Dylan-M:
### Proposed Change

The receiver read every NDJSON object twice. JSON detection failed, so the worker re-opened the object and parsed it as text lines rather than records.

The JSON parser now handles a sequence of top-level values, so NDJSON needs no detection of its own. A `json.Decoder` reads NDJSON, a lone object, and concatenated documents the same way.

A throwaway decoder classifies the peeked bytes and consumes nothing. A `json.Decoder` cannot rewind, so a `Records` search would commit the real one.

A document too large to classify inside the window uses line parsing instead, where `max_log_size` caps each record. The sequence path holds each record in memory, so the parser never buffers a very large object whole.

A malformed record now stops the object. Line parsing previously skipped it and emitted the rest as text.

`raw` still takes precedence, so a SecOps pipeline keeps the original line.

The upstream `jsonlogencodingextension` cannot serve this. Its `UnmarshalLogs` takes the whole object as a `[]byte`, so it has no streaming and no mid-object resume.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/internal/blobstream/json_parser.go
+++ b/internal/blobstream/json_parser.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"iter"
+	"strings"
 
 	"go.opentelemetry.io/collector/pdata/plog"
 	"go.uber.org/zap"
@@ -50,7 +51,7 @@ const (
 
 	// mixedSequenceLogMsg is warned once per file for a value sequence that mixes objects
 	// with non-object lines.
-	mixedSequenceLogMsg = "mixed content in JSON value sequence; string lines are emitted as bodies, malformed lines are dropped"
+	mixedSequenceLogMsg = "mixed content in JSON value sequence; text lines are emitted as bodies, corrupted JSON is dropped"
 )
 
 // MinLogSize is the smallest usable max_log_size. Content detection peeks fixed windows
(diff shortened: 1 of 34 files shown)


Dylan-M:
* **#743** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/743?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/743?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#742** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#738** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/738?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

internal/blobstream/json_parser.go
Caleb-Hurshman, line 1: Changed behavior: Mixed JSON/text files lose non-JSON lines. Previously the parser would fall back to line parsing, emitting every line without dropping data.

Caleb-Hurshman, line 1: Changed behavior: Lone object bodies and NDJSON flip from string to structured parsing. This is as intended by the feature, but it changes the output of anything downstream expecting a string body. We should make sure this change is well documented.

Caleb-Hurshman, line 1: Parsing shape is solely decided by the size of the first record. That means two files in the same format can produce different telemetry if one has a large first record and the other doesn't. At the very least, clarify in the README that the "too large to classify" designation is decided by the first record.

Caleb-Hurshman, line 1: NDJSON gets recovery that JSON doesn't: `[{"a":1},{"b":2},{bad},{"d":4},{"e":5},{"f":6}]` gets parsed correctly for NDJSON thanks to the resync, but regular JSON parsing can't recover d, e, f. This is an existing issue, but it's more noticeable now that NDJSON can handle this but JSON can't.

Caleb-Hurshman (approved):

Caleb-Hurshman (changes requested):
Sorry, accidentally approved wrong PR, still looking back at the recent changes

internal/blobstream/json_parser.go
Caleb-Hurshman: This is still around

Caleb-Hurshman: This is fixed for quoted strings, unquoted text still falls through as a parse error. Maybe acceptable for the change, but previously the mixed JSON/unquoted text would be read without parse errors

Caleb-Hurshman (approved):

Caleb-Hurshman (approved):

Dylan-M:
### Merge activity

* **Aug 26, 7:32 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/743).

PR #828, opened 2026-08-27
feat: Log type detection processor Matcher implementation

dyl10s:
### Proposed Change

Adds the matcher interface into the log type detection processor. Matchers have a priority attached to them and get run in that order. There are 2 matcher types to start
- regex
- starts with

A few additional notes:
- Log types that do not match any matchers get the `unknown` log type. There are metrics for tracking various different things. 
- The config also allows you to specify the fingerprint and log type attribute
- I will generate some documentation in a follow up PR.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/processor/logtypedetectionprocessor/config.go
+++ b/processor/logtypedetectionprocessor/config.go
@@ -15,17 +15,45 @@
 package logtypedetectionprocessor
 
 import (
+	"errors"
+
 	"go.opentelemetry.io/collector/component"
 )
 
+const (
+	defaultFingerprintField = "fingerprint"
+	defaultLogTypeField     = "log_type"
+)
+
+var errMissingLogTypeFieldError = errors.New("log_type_field is required")
+
 // Config is the config of the processor.
-type Config struct{}
+type Config struct {
+	Matchers         []MatcherConfig `mapstructure:"matchers"`
+	FingerprintField string          `mapstructure:"fingerprint_field"`
+	LogTypeField     string          `mapstructure:"log_type_field"`
+}
 
 func createDefaultConfig() component.Config {
-	return &Config{}
+	return &Config{
+		Matchers:         []MatcherConfig{},
+		FingerprintField: defaultFingerprintField,
+		LogTypeField:     defaultLogTypeField,
+	}
 }
 
 // Validate validates the processor configuration
 func (c Config) Validate() error {
+	if c.LogTypeField == "" {
+		return errMissingLogTypeFieldError
+	}
+
+	if c.Matchers != nil {
+		for _, m := range c.Matchers {
+			if err := m.Validate(); err != nil {
+				return err
+			}
+		}
+	}
 	return nil
 }
--- a/processor/logtypedetectionprocessor/config.go
+++ b/processor/logtypedetectionprocessor/config.go
@@ -23,6 +23,7 @@ import (
 const (
 	defaultFingerprintField = "fingerprint"
 	defaultLogTypeField     = "log_type"
+	unknownLogType          = "unknown"
 )
 
 var errMissingLogTypeFieldError = errors.New("log_type_field is required")
--- a/processor/logtypedetectionprocessor/config.go
+++ b/processor/logtypedetectionprocessor/config.go
@@ -26,7 +26,7 @@ const (
 	unknownLogType          = "unknown"
 )
 
-var errMissingLogTypeFieldError = errors.New("log_type_field is required")
+var errMissingLogTypeField = errors.New("log_type_field is required")
 
 // Config is the config of the processor.
 type Config struct {
@@ -46,14 +46,12 @@ func createDefaultConfig() component.Config {
 // Validate validates the processor configuration
 func (c Config) Validate() error {
 	if c.LogTypeField == "" {
-		return errMissingLogTypeFieldError
+		return errMissingLogTypeField
 	}
 
-	if c.Matchers != nil {
-		for _, m := range c.Matchers {
-			if err := m.Validate(); err != nil {
-				return err
-			}
+	for _, m := range c.Matchers {
+		if err := m.Validate(); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/processor/logtypedetectionprocessor/matcher.go
+++ b/processor/logtypedetectionprocessor/matcher.go
@@ -20,8 +20,10 @@ import (
 	"strings"
 )
 
+// MatcherType identifies how a matcher tests a log body.
 type MatcherType string
 
+// Supported matcher methods.
 const (
 	MatcherTypeRegex      MatcherType = "regex"
 	MatcherTypeStartsWith MatcherType = "starts_with"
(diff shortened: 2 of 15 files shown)


processor/logtypedetectionprocessor/metadata.yaml
```diff
@@ -18,3 +26,19 @@ telemetry:
...
+      enabled: true
+      description: "Number of logs with no detected log type."
+      stability: alpha
+      unit: "{log}"
+      sum:
+        value_type: int
+        monotonic: true
+    log_types:
+      enabled: true
+      description: "The number of hashes with a detected log type."
```
colelaven: nit for clarity
```suggestion
      description: "The number of fingerprints with a detected log type."
```

processor/logtypedetectionprocessor/matcher.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Validate validates the matcher configuration.
+func (c MatcherConfig) Validate() error {
+	if c.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+	if c.Priority < 0 {
+		return fmt.Errorf("priority must be >= 0")
+	}
+
```
colelaven: A 0 priority being allowed but then treated as "last" priority is not intuitive imo.

We should probably do one of the following:
1. Document somewhere that 0 priority means last
2. Enforce priority > 0
3. Continue to allow 0 priority, treat it as "first" priority in line with most programming things where 0 is first.

Personally I'm in favor of option 3, open to any though.

processor/logtypedetectionprocessor/matcher.go
```diff
@@ -0,0 +1,133 @@
...
+	case MatcherTypeStartsWith:
+		return newStartsWithMatcher(matcherBase, c.Value)
+	default:
+		return nil, fmt.Errorf("unknown matcher method %q", c.Method)
+	}
+}
+
+type matcherBase struct {
+	name     string
+	priority int
```
colelaven: nit - this field is never used, do we need it here?

processor/logtypedetectionprocessor/matcher.go
```diff
@@ -0,0 +1,133 @@
...
+func (m *regexMatcher) Test(s string) bool {
+	return m.regex.MatchString(s)
+}
+
+type startsWithMatcher struct {
+	matcherBase
+	prefix string
+}
+
+func newStartsWithMatcher(matcherBase matcherBase, prefix string) (*startsWithMatcher, error) {
```
colelaven: nit - err is always nil from this func we can remove the error return val

processor/logtypedetectionprocessor/metadata.yaml
```diff
@@ -18,3 +26,19 @@ telemetry:
...
         monotonic: true
+    log_type_detection_unknown:
+      enabled: true
+      description: "Number of logs with no detected log type."
+      stability: alpha
+      unit: "{log}"
+      sum:
+        value_type: int
+        monotonic: true
+    log_types:
```
colelaven: this metric name is pretty vague, could be improved to be clearer what it means. Would be nice to better clarify the difference between this and `log_type_detection_matches`

processor/logtypedetectionprocessor/metadata.yaml
```diff
@@ -10,6 +10,14 @@ status:
 
 telemetry:
   metrics:
+    log_type_detection_matches:
```
colelaven: might be nice to add a log type attribute to this metric so we can see which log types are the most often detected.

colelaven (changes requested):
I think we can improve how 0 priority is treated, and also improve the clarity of some metrics. Just some nits otherwise. Functionality looks good 👍

processor/logtypedetectionprocessor/matcher.go
dyl10s: The reason I did this goofy is `0` meant the field was unset and I wanted things with a set priority first. So I did the more smart thing and just made priority a pointer so we can handle the `nil` and `0` case separately. `nil` is now the lowest priority and `0` is first.

processor/logtypedetectionprocessor/metadata.yaml
dyl10s: I redid the metric names to make a little more sense and follow the upstream standard of `[component_type]_[component_name]_[metric]`. Let me know if they make more sense now.

processor/logtypedetectionprocessor/matcher.go
dyl10s: removed, this was going to be used for dynamic priority but it not needed for the time being.

processor/logtypedetectionprocessor/matcher.go
colelaven: Nice, the `nil` approach makes good sense to me 👍

processor/logtypedetectionprocessor/matcher.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Matcher is the compiled runtime form.
+type Matcher interface {
+	Test(s string) bool
+	Name() string
+}
+
+// Validate validates the matcher configuration.
+func (c MatcherConfig) Validate() error {
```
colelaven, line 47: I think it'd be worth adding some basic unit tests for this func

processor/logtypedetectionprocessor/config.go
```diff
@@ -15,17 +15,46 @@
...
 	"go.opentelemetry.io/collector/component"
 )
 
+const (
+	defaultFingerprintField = "fingerprint"
+	defaultLogTypeField     = "log_type"
+	unknownLogType          = "unknown"
+)
+
+var errMissingLogTypeFieldError = errors.New("log_type_field is required")
```
colelaven:
```suggestion
var errMissingLogTypeField = errors.New("log_type_field is required")
```

processor/logtypedetectionprocessor/config.go
```diff
@@ -15,17 +15,46 @@
...
+	}
 }
 
 // Validate validates the processor configuration
 func (c Config) Validate() error {
+	if c.LogTypeField == "" {
+		return errMissingLogTypeFieldError
+	}
+
+	if c.Matchers != nil {
```
colelaven: nit - redundant nil check, if it is nil the loop will just do nothing.

processor/logtypedetectionprocessor/processor.go
```diff
@@ -16,29 +16,64 @@ package logtypedetectionprocessor
...
+	p := &logTypeDetectionProcessor{
+		telemetry:        telemetry,
+		logTypeField:     cfg.LogTypeField,
+		fingerprintField: cfg.FingerprintField,
+	}
+
+	if cfg.Matchers != nil {
+		matchers := make([]MatcherConfig, len(cfg.Matchers))
+		copy(matchers, cfg.Matchers)
+		sort.SliceStable(matchers, func(i, j int) bool {
```
colelaven: Per [this package's docs](https://pkg.go.dev/sort#SliceStable):

> Note: in many situations, the newer [slices.SortStableFunc](https://pkg.go.dev/slices#SortStableFunc) function is more ergonomic and runs faster.

colelaven (changes requested):
A few more minor items, mostly g2g.

colelaven (approved):

PR #850, opened 2026-08-27
feat(restapireceiver): POST polling support

Caleb-Hurshman:
### Proposed Change

Adds a new `method` field to the REST API receiver. Valid values are `get` (default) and `post`. 
Support for POST-based polling requires the receiver to know the format that the API expects from the request body. The receiver uses the `request_body` field that takes a go template that defines where to put each value.
Example:
```yaml
method: post
request_body: |
  {
    "filter": "status:'new'",
    "limit": {{ .Limit }}{{ if .Cursor }},
    "after": "{{ .Cursor }}"{{ end }}
  }
```
This is **not** easy to configure, but the complexity is necessary to make the receiver flexible enough to handle the different nested structures that different POST-based APIs expect. Ideally, much of this complexity can be abstracted away in the source using a custom component.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


receiver/restapireceiver/factory.go
```diff
@@ -40,7 +40,11 @@ func NewFactory() receiver.Factory {
 func createDefaultConfig() component.Config {
 	return &Config{
 		URL:      "",
+		Method:   methodGET,
 		AuthMode: authModeNone,
+		// ParamLocation is deliberately left empty: Validate() derives it from the
```
Dylan-M: `ParamLocation` / `param_location` only shows up in comments (this one and `receiver.go:251`). There's no field, no config key, and `Validate()` derives nothing. Should we drop these? Actually create them? Also curious: was a `param_location: query|body` approach tried and dropped for `request_body`?

receiver/restapireceiver/client.go
```diff
@@ -95,59 +105,102 @@ func (c *defaultRESTAPIClient) Shutdown() error {
...
+
+	return body, resp.Header, nil
+}
+
+// GetJSON fetches JSON data and extracts the array of items from it.
+//
+// Not part of restAPIClient and unused by the receiver, which goes through
+// FetchFullResponse and extractDataFromResponse. Retained as the vehicle for
+// the client's authentication-mode test coverage.
+func (c *defaultRESTAPIClient) GetJSON(ctx context.Context, r apiRequest) ([]map[string]any, error) {
```
Dylan-M: `GetJSON` is dead production code, kept only for the auth-mode tests. I'd move those to `FetchFullResponse` (same `do`/`buildRequest`/`applyAuth`) and drop it.

receiver/restapireceiver/config.go
```diff
@@ -708,9 +798,32 @@ func (c *Config) validateTimestampValue(value, fieldName string) error {
...
+		return fmt.Errorf("%s %q is not a valid JSON number; an epoch timestamp must be plain digits with no leading \"+\" and no trailing \".\"", fieldName, value)
+	}
+
 	return nil
 }
 
+// isValidJSONNumber reports whether s can be emitted as a bare JSON number.
+// json.Number carries its contents verbatim without validating them, so an
+// illegal value would otherwise fail at marshal time rather than config time.
+func isValidJSONNumber(s string) bool {
```
Dylan-M, line 781: `isValidJSONNumber` hand-rolls the number check. Let's decode into `json.Number` instead of maintaining our own.

receiver/restapireceiver/requestbody.go
```diff
@@ -0,0 +1,95 @@
...
+
+// parseRequestBodyTemplate parses a request_body template. The template renders
+// against requestBodyData, so a reference to a field that does not exist fails
+// when it is executed rather than here.
+func parseRequestBodyTemplate(body string) (*template.Template, error) {
+	return template.New(requestBodyTemplateName).Parse(body)
+}
+
+// renderRequestBody renders the template for one request.
+func renderRequestBody(tmpl *template.Template, data requestBodyData) ([]byte, error) {
```
Dylan-M, line 63: Optional: a `{{ json .Cursor }}` helper would let people skip hand-quoting strings and drop the "unquoted string value" case. Numbers like `{{ .Limit }}` stay bare.

receiver/restapireceiver/config.go
```diff
@@ -475,6 +531,26 @@ func (c *Config) Validate() error {
...
+	if c.Method == "" {
+		c.Method = methodGET
+	}
+
+	// Validate method
+	switch c.Method {
+	case methodGET, methodPOST:
+		// Valid methods
+	default:
+		return fmt.Errorf("invalid method: %s, must be one of: get, post", c.Method)
```
Dylan-M: nit: `validate()` re-checks the method `UnmarshalText` already rejects, same error string. This could be a single check.

receiver/restapireceiver/pagination.go
```diff
@@ -586,3 +549,101 @@ func checkPageLimit(cfg *Config, state *paginationState) bool {
...
+		PageSize:  state.PageSize,
+		StartTime: jsonEscapeString(requestStartTime(cfg, state)),
+		EndTime:   jsonEscapeString(state.ResolvedEndTime),
+	}
+}
+
+// jsonEscapeString escapes s for use inside a JSON string literal, without the
+// surrounding quotes. Epoch and RFC3339 timestamps pass through unchanged, so a
+// template may also use them unquoted as bare JSON numbers.
+func jsonEscapeString(s string) string {
```
Dylan-M: nit: the error branch in `jsonEscapeString` is dead. `json.Marshal` of a string never errors.

receiver/restapireceiver/receiver.go
```diff
@@ -519,29 +604,35 @@ func (r *restAPILogsReceiver) poll(ctx context.Context) (pollResult, error) {
 
 	result := pollResult{}
 
-	// Build initial pagination parameters
-	params := buildPaginationParams(r.cfg, r.paginationState)
+	// Build the initial request from the current pagination state
+	req, err := r.buildAPIRequest()
```
Dylan-M, line 608: nit: the logs and metrics `poll()` loops are near-identical. Pre-existing, but this is more tech debt that could be refactored into more maintainable code.

Dylan-M (changes requested):
A few things inline.

receiver/restapireceiver/factory.go
Caleb-Hurshman: Yeah, that was an earlier iteration that I (mostly) reverted.

receiver/restapireceiver/requestbody.go
Caleb-Hurshman: Didn't go this route because requestBodyData fields are pre-escaped, so a JSON helper would double-escape. I'd consider this, but hopefully most users are configuring using the source anyway.

Dylan-M: Right, on top of the pre-escaping it'd double-escape. I meant replacing it: `newRequestBodyData` stops pre-escaping and `{{ json }}` escapes at the call site. Whether people configure through the source or not, one escaping site reads cleaner than pre-escaping every field.

receiver/restapireceiver/config.go
Caleb-Hurshman: Removed all 5 redundant error checks from validate

Dylan-M (approved):

PR #863, opened 2026-09-01
fix: Add configurable size for page size pagination (PIPE-1442)

Caleb-Hurshman:
### Proposed Change

Adds a `page_size` parameter for the Page/Size pagination mode. Not sure how this ended up this way, but the size was always set to 20 with no way to configure it for that specific pagination mode.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


eKuG (approved):

Dylan-M (changes requested):
The new `page_size < 0` check has no test, though its mirror `limit < 0` does (the "negative offset_limit limit" case in config_test.go). Please add the matching negative-page_size case so the branch is covered.

Dylan-M (approved):

PR #901, opened 2026-09-02
feat: Fingerprint storage extension for log type detection

dyl10s:
### Proposed Change

Adds configuration options for a storage extension to be used to save and load the fingerprint map.

Multiple options are configurable for how the fingerprint map is persisted to the storage extension:
- `fingerprint_persist_interval` - defines how often the fingerprint map should be persisted to disk the map is always persisted on exit
- `max_saved_fingerprints` - defines the maximum number of fingerprints to save to disk before the oldest fingerprints are removed

Some other changes that I did:
- Updated the helper for creating a storage extension to have a kind parameter
- Fingerprint map invalidates if matchers change
- Switched to using an LRU map for the hash table

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


processor/logtypedetectionprocessor/processor.go
```diff
@@ -76,15 +132,103 @@ func priorityRank(priority *int) int {
...
+		p.id,
+		pipeline.SignalLogs,
+	)
+	if err != nil {
+		return fmt.Errorf("create storage client: %w", err)
+	}
+	saved := persistedFingerprints{}
+	if err := client.LoadStorageData(ctx, fingerprintStorageKey, &saved); err != nil {
+		return errors.Join(fmt.Errorf("load log types: %w", err), client.Close(ctx))
+	}
```
colelaven, line 156: if there is an error loading the storage data, do you think it'd make sense to just log an error / warn, and ignore the current storage data? Kinda similar to line 159.

processor/logtypedetectionprocessor/go.mod
```diff
@@ -37,18 +45,21 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
+	github.com/observiq/bindplane-otel-contrib/internal/storageclient v1.13.0
```
colelaven:
```suggestion
	github.com/observiq/bindplane-otel-contrib/internal/storageclient v1.14.0
```

processor/logtypedetectionprocessor/processor.go
```diff
@@ -76,15 +132,103 @@ func priorityRank(priority *int) int {
...
+
+	ticker := time.NewTicker(p.cfg.FingerprintPersistInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if err := p.save(ctx); err != nil {
```
colelaven, line 196: optimization nit - Some indication of if any new log types have been added since the last `save` would allow us to skip re-writing the storage data if it hasn't changed in the last tick period.

colelaven (changes requested):
some nits, mostly looking good.

1 case to call out, you may have already considered it. 

If the fingerprinting logic is changed, but the matchers do not (so basically on a collector upgrade), loading the stored `persistedFingerprints` state could contain a lot of stale fingerprint data that won't be relevant anymore (and could potentially collide with new fingerprints? probably unlikely though). As long as collisions like that aren't likely, its probably not a huge deal, the LRU caching will overwrite stale data eventually.

processor/logtypedetectionprocessor/processor.go
dyl10s: My concern here would be the edge case where the storage is on some remote system or even some local case, where we fail to read it and start building a fresh fingerprint list and then at some future point we have access to the storage again and the whole list gets overwritten.

dyl10s:
> some nits, mostly looking good.
> 
> 1 case to call out, you may have already considered it.
> 
> If the fingerprinting logic is changed, but the matchers do not (so basically on a collector upgrade), loading the stored `persistedFingerprints` state could contain a lot of stale fingerprint data that won't be relevant anymore (and could potentially collide with new fingerprints? probably unlikely though). As long as collisions like that aren't likely, its probably not a huge deal, the LRU caching will overwrite stale data eventually.

In this case the hashes would be different and the cache would slowly get purged. The likelihood of a duplicate would be the same likelihood of 2 different log structures having a duplicate hash.

processor/logtypedetectionprocessor/processor.go
dyl10s: good idea 💯

processor/logtypedetectionprocessor/processor.go
colelaven: Makes sense 👍

colelaven (approved):

PR #941, opened 2026-09-07
fix(lookupprocessor): start with an unreachable Redis and bound the cost of a source outage (PIPE-1450)

eKuG:
### Proposed Change

The Redis lookup source ran a `PING` in its constructor and returned an error when it failed. That error propagated out of the processor's `Start`, so a Redis that was down when the collector started (restart, network blip, maintenance window) failed the whole collector, not just enrichment. Enrichment is best-effort: the API source never probes its endpoint at start, per-record lookup failures already pass the record through un-enriched, and a missing CSV file is a background error rather than a start failure. Redis was the only source that could take the pipeline offline.

Changes:

- **Startup is fail-soft.** The boot `PING` stays as a probe, but a failure logs one `warn` and the source starts anyway. go-redis dials lazily on the next command, so Redis is picked up when it comes back with no collector restart. Startup now behaves the same as the API source.
- **Outage transitions are logged once.** A small `availability` tracker shared by the redis and api sources logs one `warn` when the source stops answering and one `info` when it answers again. Per-record failures stay at `debug`. A 4xx from the API (including 404) is a per-key miss, not an outage.

### Describe How The Reviewer Can Validate The Changes

1. `cd processor/lookupprocessor && go test -race ./...`
2. Build a collector from this branch (`make build-linux-arm64` with `COLLECTOR_PATH` pointing at a collector checkout) and run it with a `lookup` processor whose `redis.address` points at a host that is not listening. Before this change the collector exits 1 with `redis ping failed`; after it, the collector logs `redis unreachable at startup` and reaches `Everything is ready`.
3. Send a record through; it arrives un-enriched. Start Redis on that address and seed a key; within 5s the next record is enriched and the log shows `redis lookup source reachable again`.
4. `docker pause` the Redis container and send several records: only the first blocks for `lookup_timeout`, the rest return immediately.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed

### DIFF
--- a/processor/lookupprocessor/availability_test.go
+++ b/processor/lookupprocessor/availability_test.go
@@ -31,10 +31,10 @@ import (
 func fakeAvailability(t *testing.T) (*availability, *clockwork.FakeClock, *observer.ObservedLogs) {
 	t.Helper()
 	core, logs := observer.New(zapcore.DebugLevel)
-	clock := clockwork.NewFakeClock()
+	fakeClock := clockwork.NewFakeClock()
 	a := newAvailability(zap.New(core), "test")
-	a.clock = clock
-	return a, clock, logs
+	a.clock = fakeClock
+	return a, fakeClock, logs
 }
 
 func TestAvailability_UpNeverSkips(t *testing.T) {
@@ -47,7 +47,7 @@ func TestAvailability_UpNeverSkips(t *testing.T) {
 }
 
 func TestAvailability_DownSkipsUntilWindowThenOneProbe(t *testing.T) {
-	a, clock, logs := fakeAvailability(t)
+	a, fakeClock, logs := fakeAvailability(t)
 	boom := errors.New("boom")
 
 	a.markDown(boom)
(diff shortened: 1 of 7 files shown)


processor/lookupprocessor/redis.go
```diff
@@ -123,6 +134,7 @@ func (r *RedisSource) Lookup(ctx context.Context, key string) (map[string]string
 		if errors.Is(err, redis.Nil) {
 			return nil, errRedisKeyNotFound
 		}
+		r.avail.markDown(err)
```
Dylan-M, line 137: This isn't covered. The outage tests kill the server, so `HGetAll` fails first and returns before GET.

processor/lookupprocessor/availability_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func fakeAvailability(t *testing.T) (*availability, *clockwork.FakeClock, *observer.ObservedLogs) {
+	t.Helper()
+	core, logs := observer.New(zapcore.DebugLevel)
+	clock := clockwork.NewFakeClock()
```
Dylan-M: You're using `clock` as the variable name for a fake clock, and this could be confused with the real wall clock. I suggest either `fclock` or `fakeclock`.

Dylan-M (changes requested):

Dylan-M (approved):

eKuG:
### Merge activity

* **Sep 10, 4:56 PM UTC**: [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/941) rebased this pull request as part of a merge.
* **Sep 10, 5:21 PM UTC**: @eKuG merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/941).

PR #942, opened 2026-09-08
chore(restapireceiver): Refactor logs and metrics polling loops

Caleb-Hurshman:
### Proposed Change

Before, the REST API receiver had two separate poll loops for logs and metrics collection. This PR consolidates them to make the code more maintainable.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Dylan-M (approved):
Non-blocking: the consume-error branches in `consumeLogs` and `consumeMetrics` aren't covered. They were untested in each of the old duplicated polls too, so nothing regresses here.

PR #943, opened 2026-09-08
feat(restapireceiver): Add `has_more_field_name` field

Caleb-Hurshman:
### Proposed Change

Previously, the REST receiver relied on the `limit` field to determine whether an API had returned a full response and had more telemetry to query. Some APIs operate this way, but others will include a field that explicitly says if there is more information to query.
This PR adds a `pagination.has_more_field_name` field to the receiver's configuration to simplify the logic and enable the `limit` field to act as a throughput control.

##### Checklist
- [ ] Changes are tested
- [ ] CI has passed


Dylan-M (approved):
Non-blocking: `explicitHasMore`'s non-map fallback isn't tested. The same fallback in `parseOffsetLimitResponse` has a test, so this is the only place it's untested.

PR #965, opened 2026-09-16
fix(azureblobpollingreceiver): add a streaming mode for blobs that grow via append

Dylan-M:
### Proposed Change
Adds streaming support for blobs that grow via append block while preserving the existing whole blob behavior in parallel.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed

PR #966, opened 2026-09-16
feat(azureblobpollingreceiver): incremental byte-offset reads for append-growable blobs

Dylan-M:
### Proposed Change
Adds opt-in incremental byte-offset reads (`enable_incremental_read`) for the `records-json`
and `json` formats. Also adds fingerprinting so that any blob that gets replaced gets properly detected as new (mirrors file log receiver as much as possible).

Defaults to off, legacy whole-blob behavior unchanged.

##### Checklist
- [X] Changes are tested
- [ ] CI has passed


Dylan-M:
* **#968** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/968?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#967** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#966** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/966?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/966?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#965** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/965?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

receiver/azureblobpollingreceiver/receiver.go
```diff
@@ -674,6 +861,340 @@ func (r *pollingReceiver) processBlobGoRoutine(ctx context.Context, blob *azureb
...
+			}
+			wrapped := fmt.Errorf("download blob range: %w", err)
+			if offset == 0 {
+				r.trackNeverReadFailure(blob, wrapped)
+			}
+			return false, wrapped
+		}
+		// Offset at/past the blob end: no new bytes. Fall through with an empty delta
+		// so LastModified still advances and the mtime gate stops re-firing this read.
+		delta = nil
```
eKuG, line 1116: A 416 here only proves size <= offset, and this branch assumes equality (no growth). If the blob was replaced by one that shares the identical first fingerprint-size bytes but is shorter than the stored offset (fpSize <= newSize < offset), verifyResumeIdentity passes, this branch parks the offset, and the replacement's content past the shared prefix is skipped; once it regrows past the old offset, reads resume mid-record. SameBlob's own doc says shrunk = replaced, so this path contradicts that doctrine for the one case the fingerprint can't see.

Listed blob.Size can't disambiguate (stale by design), but one authoritative size lookup on this branch can: GetBlobProperties, or the Content-Range total from the verify head-read. size < offset then means shrunk, so reset to 0 like any other replacement. Cheap because this branch only fires on no-growth polls.

Requires offset > fingerprint_size and an identical 512-byte prefix on the replacement, so it's a narrow window; flagging rather than blocking.

eKuG (commented):
One edge in the 416 path worth closing; everything else in the offset/fingerprint machinery checked out (commit-after-consume ordering, committed-prefix fingerprinting, torn-record splits, map locking).

eKuG (approved):

Dylan-M:
### Merge activity

* **Sep 16, 3:15 PM UTC**: @Dylan-M merged this pull request with [Graphite](https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/966).

PR #969, opened 2026-09-16
fix(snapshotprocessor): bound snapshot buffer copies and rate-limit admission (BPOP-5844)

eKuG:
### Proposed Change
Bindplane inserts three snapshot processors into every pipeline, and each one deep-copied every batch before the shared buffer in `pkg/snapshot` kept only 100 of its records. On a busy pipeline that copy is the dominant allocation source of the processor (BPOP-5844, profiles from a customer collector).

This reworks `pkg/snapshot` so the cost no longer scales with throughput:

- `Add` copies only the newest records the buffer will keep (at most the ideal size) into a single record store and evicts the oldest in place. There is no list of batches to re-count, grow or pin.
- Once a buffer is full it admits about `idealSize` items per `DefaultRefreshInterval` (1 s), budgeted by records so batch size does not matter; below capacity everything is admitted. Rejected payloads are never read. `WithRefreshInterval(0)` restores admit-everything.
- `ConstructPayload` copies the buffered records under the lock and does filtering, sampling, marshaling and compression unlocked, in one pass instead of five. gzip writers are pooled.
- Zero-count payloads are dropped, `Len` is O(1) and synchronized, empty resource/scope/metric groups are never copied, the sampler bounds its index, and the compressed size is always measured before a payload is accepted.
- The processor hands batches straight to `Add`. Benchmarks cover the hot path and the request path.

The public API of `pkg/snapshot` is unchanged. Behaviour changes: a batch larger than the ideal size retains its newest records instead of the whole batch (traces golden updated), and a full buffer refreshes about 100 records per second rather than on every batch.

Measured: hot path 227 µs / 13,012 allocs → 78 ns / 0 allocs per 1000-record batch per processor; snapshot request 2.0 ms / 6.0 MB → 0.9 ms / 0.5 MB. In Docker with three processors at ~13k log records/s against the same pipeline without them, allocation overhead went from 25% to 2.6% (1-record batches) and from 20% to 0.1% (1000-record batches); `processLogs` from 9.4% to 1.1% of CPU samples. Verified end to end against a local Bindplane: source, destination and agent snapshots return records and search works.

Consumed by observIQ/bindplane-otel-collector#3648.

#### On-demand collection for high-throughput pipelines

Added after review: at every window rollover the buffer takes the record and batch rate of the window it just closed (every offered payload is counted). After three consecutive windows above ten buffers' worth of records per second with at least ten batches per second, the buffer drops its store and stops collecting; `Add` then costs one atomic load. A snapshot request arms the buffer, drops whatever the store held, waits until the store is full or one second has passed, answers, and the last in-flight request disarms it and drops the store again. A request whose wait times out returns the buffer to continuous collection, keeping the partial records. `WithRefreshInterval(0)` disables detection along with the budget. No API change.

Benchmarks were added first (`BenchmarkLogBufferFreshness`, `BenchmarkLogBufferRetainedHeap`) to establish the baseline. Results (Apple M-series): idle hot path 227 µs / 13k allocs (stock) → 53 ns (continuous) → ~2 ns / 0 allocs (on-demand); retained heap per full buffer 72 KB → 0.2 KB; age of returned records at 2k and 20k records/s up to 1 s → 1–2 ms; request latency becomes the time to collect 100 records (50 ms at 2k/s, 5 ms at 20k/s). A 200 records/s pipeline stays continuous. Docker A/B with three snapshot processors at ~13k records/s: the snapshot path is 0.21% of CPU samples at 1-record batches and absent at 1000-record batches; allocation overhead +2.6% and +0.0%. Browser E2E with two agents on one Bindplane: the 2,006 records/s pipeline returns a single ~50 ms just-in-time slice per request (records 17–19 ms old, five refreshes strictly advancing); the 20 records/s pipeline stays continuous and serves its history at once. Full write-up with every command: `pkg/snapshot/PERFORMANCE.md`.

Known and documented: a request that begins in the microsecond the buffer flips can copy an empty store without waiting (the next poll a second later self-heals); on-demand polls return a sample per poll rather than a rolling tail (continuous mode also admits ~100 records/s, so volume per poll is unchanged); the legacy report-manager path waits up to 1 s inside the OpAMP reload callback, where it already blocked on an HTTP POST.

GCP VM (x86-64, n2-standard-8, collectors capped at 2 CPUs), same rig: stock +27.7% CPU / +25.1% allocation at 1-record batches and +5.2% / +20.0% at 1000-record batches; bounded-copy build +6.2% / +2.7% and +0.2% / +0.2%; on-demand build +7.0% / +2.4% and +0.1% / +0.2%, snapshot path 0.34% and 0.049% of CPU samples. The rig (compose, configs, run script, Dockerfile, VM runbook) is checked in under `pkg/snapshot/testdata/perfrig`.

##### Checklist
- [x] Changes are tested
- [ ] CI has passed


eKuG:
* **#969** <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/969?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/observIQ/bindplane-otel-contrib/969?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pkg/snapshot/admission.go
```diff
@@ -0,0 +1,295 @@
...
+	// enterOnDemand ignores the payload, drops the store and stops
+	// collecting: the pipeline is fast enough for just-in-time collection.
+	enterOnDemand
+)
+
+// init sets the interval and budget and opens the first window now.
+func (a *admission) init(interval time.Duration, budget int) {
+	a.interval = interval
+	a.budget = int64(budget)
+	a.windowNs.Store(time.Now().UnixNano())
```
jsirianni: We should use the Time package's monotonic clock capability when measuring time changes. See https://pkg.go.dev/time#hdr-Monotonic_Clocks

e.g.

```go
var epoch = time.Now() // carries a monotonic reading
func monoNow() int64 { return int64(time.Since(epoch)) }
```

I used Claude to find this. Here is the raw findings.

> A backward clock step freezes the buffer (new bug, low-to-medium)

Where: pkg/snapshot/admission.go:127, :150-153, :236

The admission window uses time.Now().UnixNano(), which is wall-clock time. If the system clock steps backwards by some amount Δ (NTP step, VM restore, a manual change), now-start goes negative. Once the ~100-record budget for the window is spent, decide rejects every batch until the wall clock catches up, which takes Δ + 1s.

A busy pipeline spends its budget almost immediately, so continuous-mode snapshots keep serving the same old records for as long as the clock was stepped back. The pipeline's own data is unaffected; only the snapshot view goes stale.

I confirmed it with a probe: I set windowNs one hour ahead, waited 1.1s and added 50 new batches. 0 were admitted.

Fix: measure against the monotonic clock.
var epoch = time.Now() // carries a monotonic reading
func monoNow() int64 { return int64(time.Since(epoch)) }
Use monoNow() in init, decide and beginRequest. sampleWindow already guards durationNs <= 0.

pkg/snapshot/compress.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// compressedSize returns the size data would gzip compress to, without
+// allocating or retaining the compressed bytes. It is safe for concurrent
+// use.
+func compressedSize(data []byte) (int, error) {
+	cw := &countingWriter{}
+
+	w := gzipWriterPool.Get().(*gzip.Writer)
+	defer gzipWriterPool.Put(w)
```
jsirianni: Call w.Reset in the defer, before Put. Otherwise the last payload is kept in memory until the buffer is used again or GC. 

```suggestion
	defer func() { 
	  w.Reset(io.Discard)
	  gzipWriterPool.Put(w) 
	}()
```

jsirianni (changes requested):
@eKuG  Do we expect the snapshot processor to return the last known snapshot, if the telemetry volume drops to 0? That is today's behavior, and can be useful. I believe this new mode would return an empty snapshot if telemetry volume drops to 0.

eKuG:
@jsirianni you're right. I fixed it.
On-demand mode used to drop the buffer when it switched modes and again after each request, so a request after traffic stopped came back empty. 
The buffer keeps its last snapshot, adds one heartbeat batch every 10 s when idle, and waits for a full buffer of fresh records before sending requests. If the pipeline is running, it sends records at the time of the request. If the pipeline is stopped, it sends the last known snapshot, at most 10 seconds older than when it stopped, and the buffer continues to collect.

